PR #2, opened 2022-06-23
Feat/aave pools helpers

eboadom:


src/test/AaveV3Helpers.t.sol
```diff
@@ -0,0 +1,101 @@
...
+
+    function setUp() public {
+        startHoax(USER_1);
+    }
+
+    function testDepositV3() public {
+        address depositor = USER_1;
+        address onBehalf = depositor;
+        address asset = AaveAddressBookV3
+            .getToken(MARKET_NAME, "DAI.e")
```
sakulstra, line 21: this will not work when MARKET_NAME is different i guess :thinking: 

still not sure if this `getToken(symbol)` makes any sense at all

sakulstra: the usefulness of `getToken` is limited so we removed it from address-book meaning this code needs to change a bit

sakulstra (changes requested):

sakulstra:
can be closed i think

PR #12, opened 2022-09-16
feat: more test utils

sakulstra:


src/ProtocolV3TestBase.sol
```diff
@@ -448,4 +479,326 @@ contract ProtocolV3TestBase is Test {
...
+    require(allConfigsBefore.length == allConfigsAfter.length, 'A_UNEXPECTED_NEW_LISTING_HAPPENED');
+
+    for (uint256 i = 0; i < allConfigsBefore.length; i++) {
+      if (assetChangedUnderlying != allConfigsBefore[i].underlying) {
+        _requireNoChangeInConfigs(allConfigsBefore[i], allConfigsAfter[i]);
+      }
+    }
+  }
+
+  function _requireNoChangeInConfigs(ReserveConfig memory config1, ReserveConfig memory config2)
```
eboadom, line 675: Probably because it was not good before, but I think some configs are not validated on this function

src/ProtocolV3TestBase.sol
```diff
@@ -448,4 +479,326 @@ contract ProtocolV3TestBase is Test {
...
+    ReserveConfig[] memory allConfigsAfter
+  ) internal pure {
+    require(
+      allConfigsBefore.length == allConfigsAfter.length - count,
+      '_validateCountOfListings() : INVALID_COUNT_OF_LISTINGS'
+    );
+  }
+
+  function _validateReserveTokensImpls(
+    address poolConfigurator,
```
eboadom: Should we pass always addresses provider and using the getter here? Or is it to be compatible with v2 and v3?

src/ProtocolV3TestBase.sol
sakulstra: good catch - didn't notice, added all the remaining :+1:

src/ProtocolV3TestBase.sol
sakulstra: nah, just didn't think about it - probably makes sense to stay consistent. refactored to addressProvider

eboadom (approved):

PR #14, opened 2022-09-21
feat: add repay method

sakulstra:


src/ProtocolV3TestBase.sol
```diff
@@ -201,11 +201,23 @@ contract ProtocolV3TestBase is Test {
     IPool pool,
     uint256 amount,
     bool stable
-  ) internal {
+  ) external {
```
sakulstra, line 217: made it external as otherwise we cannot try catch it and check specific errors

src/ProtocolV3TestBase.sol
```diff
@@ -201,11 +201,23 @@ contract ProtocolV3TestBase is Test {
...
     console.log('BORROW: %s, Amount %s, Stable: %s', config.symbol, amount, stable);
     pool.borrow(config.underlying, amount, stable ? 1 : 2, 0, address(this));
   }
 
+  function _repay(
```
sakulstra, line 226: wdyt, in most current helpers we require the user to have funds before and provide a flag for approval

I don't see where this is useful as we are not testing the protocol, so having funds and approval could be implicit.

sendra: yup I fully agree with this. having funds and approved should be defaulted on the method.

src/ProtocolV3TestBase.sol
```diff
@@ -83,7 +83,7 @@ contract ProtocolV3TestBase is Test {
 
   function e2eTest(IPool pool) public {
     ReserveConfig[] memory configs = _getReservesConfigs(pool);
-    deal(address(this), 1000 ether);
+    deal(msg.sender, 1000 ether);
```
sakulstra, line 86: makes more sense to always use msg.sender as that way you can hoax / prank whoever you want

eboadom: Actually, what is `msg.sender` on the context of a forge test?

sakulstra: the current address which you can alter via hoax and prank ... i think if you don't alter it's the test contract

src/ProtocolV3TestBase.sol
eboadom: Yes, on previous helpers I made, some stuff was clearly useless

PR #20, opened 2022-12-12
Add GenericV3ListingEngine

eboadom:


src/interfaces/IERC20.sol
```diff
@@ -0,0 +1,82 @@
+// SPDX-License-Identifier: MIT
```
sendra, line 1: This could be imported from https://github.com/bgd-labs/solidity-utils/blob/main/src/contracts/oz-common/interfaces/IERC20.sol

sendra:
https://github.com/bgd-labs/aave-helpers/blob/eebcded1684bb0feed13fb01f4d9bd9fd42c0618/src/test/ProxyHelpers.t.sol#L10 this needs to be updated to 'mainnet' fork

src/interfaces/IERC20.sol
eboadom: Not including `decimals()` the OZ version, that is why I use another one

sendra (approved):

PR #25, opened 2023-01-16
GenericV3ListingEngine v0.1 final version

eboadom:
Version of the engine to be used for the initial Aave v3 Ethereum payload.

Only exposing `listAssets()` externally, as the engine is not well prepared yet for more granular updates.

### DIFF
--- a/src/v3-listing-engine/GenericV3ListingEngine.sol
+++ b/src/v3-listing-engine/GenericV3ListingEngine.sol
@@ -8,6 +8,7 @@ import {IGenericV3ListingEngine} from './IGenericV3ListingEngine.sol';
 
 /**
  * @dev Helper smart contract implementing a generalized Aave v3 listing flow for a set of assets
+ * Version 0.1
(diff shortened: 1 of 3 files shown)


src/v3-listing-engine/GenericV3ListingEngine.sol
```diff
@@ -8,6 +8,7 @@ import {IGenericV3ListingEngine} from './IGenericV3ListingEngine.sol';
 
 /**
  * @dev Helper smart contract implementing a generalized Aave v3 listing flow for a set of assets
+ * Version 0.1
```
sakulstra: i see the motivation, but not a fan of versioning here as it will probably mismatch with the overall package version.

sakulstra (approved):

src/v3-listing-engine/GenericV3ListingEngine.sol
eboadom: Maybe we can boil it down to the other versioning if documented appropriately. Will remove

PR #29, opened 2023-01-19
Added TestWithExecutor contract to GovHelpers

eboadom:
It is pretty common use case across other projects to have some payload execution infra for tests, so seems appropriate to add it here

### DIFF
--- a/src/GovHelpers.sol
+++ b/src/GovHelpers.sol
@@ -92,21 +92,21 @@ contract MockExecutor {
 /**
  * @dev Inheriting from this contract in a forge test allows to
  * 1. Configure on the setUp() of the child contract an executor for governance proposals
- *    (or any address with permissions) just by doing for example a `_setUp(AaveGovernanceV2.SHORT_EXECUTOR)`
- * 2. Afterwards, on a test you can just do `_execute(somePayloadAddress)`, and it will be executed via
+ *    (or any address with permissions) just by doing for example a `_selectPayloadExecutor(AaveGovernanceV2.SHORT_EXECUTOR)`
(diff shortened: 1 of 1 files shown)


src/GovHelpers.sol
```diff
@@ -73,3 +77,36 @@ library GovHelpers {
     return GOV.getProposalById(proposalId);
   }
 }
+
+/**
+ * @dev Mock contract which allows performing a delegatecall to `execute`
+ * Intended to me used as replacement for L2 admins to mock governance/gnosis execution.
```
sakulstra:
```suggestion
 * Intended to be used as replacement for L2 admins to mock governance/gnosis execution.
```

src/GovHelpers.sol
```diff
@@ -73,3 +77,36 @@ library GovHelpers {
...
+ * @dev Inheriting from this contract in a forge test allows to
+ * 1. Configure on the setUp() of the child contract an executor for governance proposals
+ *    (or any address with permissions) just by doing for example a `_setUp(AaveGovernanceV2.SHORT_EXECUTOR)`
+ * 2. Afterwards, on a test you can just do `_execute(somePayloadAddress)`, and it will be executed via
+ *    DELEGATECALL on the address previously selected on step 1).
+ */
+abstract contract TestWithExecutor is Test {
+  MockExecutor internal _executor;
+
+  function _setUp(address executor) internal {
```
sakulstra: _setUp seems a little to generic for being in the library - i fear this will create name clashes.
Perhaps better _deployMockExecutor or sth along these lines

eboadom: `_selectPayloadExecutor()`? To follow "select" wording of foundry

sakulstra: Okay

PR #40, opened 2023-01-31
feat: v2 test base

sakulstra:


src/ProtocolV2TestBase.sol
```diff
@@ -0,0 +1,852 @@
...
+        );
+        vm.writeLine(
+          path,
+          string(
+            abi.encodePacked(
+              abi.encodePacked(
+                '| ',
+                vm.toString(address(strategy)),
+                ' | ',
+                'TODO',
```
sakulstra: left a `TODO` here as idk yet how i can dynamically fetch the stable rate oracle per pool

eboadom:
Should the base stable rate just be added on the specific of the rate strategy, even if in v2 fetched from the lending rate oracle? https://github.com/bgd-labs/aave-helpers/blob/e0a0f7bbe802700678654705d8ae0ec36c87dc2e/src/ProtocolV2TestBase.sol#L400

eboadom:
Should we have only the necessary functions? Or import from somewhere else maybe? https://github.com/bgd-labs/aave-helpers/blob/e0a0f7bbe802700678654705d8ae0ec36c87dc2e/src/ProtocolV2TestBase.sol#L96

eboadom:
A typo here (actually on v3 too) https://github.com/bgd-labs/aave-helpers/blob/e0a0f7bbe802700678654705d8ae0ec36c87dc2e/src/ProtocolV2TestBase.sol#L267

eboadom:
Should we change to the helper function of forge-std on this? https://github.com/bgd-labs/aave-helpers/blob/e0a0f7bbe802700678654705d8ae0ec36c87dc2e/src/ProtocolV2TestBase.sol#L861

src/ProtocolV2TestBase.sol
eboadom: Mmh, just `getLendingRateOracle()` on the addresses provider or is there any problem?

sakulstra: The comment was already outdated

sakulstra:
@eboadom 

> Should the base stable rate just be added on the specific of the rate strategy, even if in v2 fetched from the lending rate oracle?

I think this doesn't work as the baserate is per asset, not per strategy.

> Should we have only the necessary functions? Or import from somewhere else maybe?

Moved it to interfaces for now, so at least it's not duplicated. might make sense in solidity utils though.

> A typo here (actually on v3 too)

fixed

> Should we change to the helper function of forge-std on this?

changed

eboadom (approved):

PR #42, opened 2023-02-01
!feat: proposal creation helpers

sakulstra:
BREAKING:
- GovHelpers no longer exports any constant addresses
- `GovHelpers.createProposal` was renamed to `GovHelpers.createProposalAsWhale`

### DIFF
--- a/src/GovHelpers.sol
+++ b/src/GovHelpers.sol
@@ -6,9 +6,20 @@ import {Vm} from 'forge-std/Vm.sol';
 import {Test} from 'forge-std/Test.sol';
 import {AaveGovernanceV2, IAaveGovernanceV2, IExecutorWithTimelock} from 'aave-address-book/AaveGovernanceV2.sol';
 import {IPoolAddressesProvider} from 'aave-address-book/AaveV3.sol';
+import {AaveMisc} from 'aave-address-book/AaveMisc.sol';
 import {ProxyHelpers} from './ProxyHelpers.sol';
 
-library DelegatecallProposalCreationHelper {
+library GovHelpers {
+  struct SPropCreateParams {
+    address executor;
+    address[] targets;
+    uint256[] values;
+    string[] signatures;
+    bytes[] calldatas;
+    bool[] withDelegatecalls;
+    bytes32 ipfsHash;
+  }
+
   struct DelegateCallProposal {
     address target;
     string signature;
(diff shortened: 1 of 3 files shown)


src/GovHelpers.sol
```diff
@@ -8,6 +8,86 @@ import {AaveGovernanceV2, IAaveGovernanceV2, IExecutorWithTimelock} from 'aave-a
 import {IPoolAddressesProvider} from 'aave-address-book/AaveV3.sol';
 import {ProxyHelpers} from './ProxyHelpers.sol';
 
+library DelegatecallProposalCreationHelper {
```
sakulstra: Should this even be a special library or part of `GovHelpers` ?

src/GovHelpers.sol
```diff
@@ -8,6 +8,86 @@ import {AaveGovernanceV2, IAaveGovernanceV2, IExecutorWithTimelock} from 'aave-a
...
+  struct DelegateCallProposal {
+    address target;
+    string signature;
+    bytes callData;
+  }
+
+  function createMainnetDelegateCall(address payloadAddress)
```
sakulstra: not sure about the naming

src/GovHelpers.sol
```diff
@@ -8,6 +8,86 @@ import {AaveGovernanceV2, IAaveGovernanceV2, IExecutorWithTimelock} from 'aave-a
...
+  {
+    return
+      DelegateCallProposal({
+        target: AaveGovernanceV2.CROSSCHAIN_FORWARDER_POLYGON,
+        signature: 'execute(address)',
+        callData: abi.encode(payloadAddress)
+      });
+  }
+
+  function createProposal(DelegateCallProposal[] memory delegateCalls, bytes32 ipfsHash)
```
sakulstra: we already have `createProposal` in `GovHelpers` - the main difference beeing that it automatically mocks a whale etc, so it's not usable in breadcast. Does it make sense to rename the mock one to `createProposalViaPrankedWhale` or similar?

src/GovHelpers.sol
eboadom: Why not `createMainnet()`, you already have the other on the Lib name

sendra (approved):

src/GovHelpers.sol
```diff
@@ -19,24 +20,93 @@ library GovHelpers {
...
+      targets[i] = delegateCalls[i].target;
+      signatures[i] = delegateCalls[i].signature;
+      calldatas[i] = delegateCalls[i].callData;
+      values[i] = 0;
+      withDelegatecalls[i] = true;
+    }
+
+    return
+      AaveGovernanceV2.GOV.create(
+        IExecutorWithTimelock(AaveGovernanceV2.SHORT_EXECUTOR),
```
eboadom, line 96: We should remember that this is biased for short exec and not screw on some long exec proposal

sakulstra: we might add an overload allowing to pass an executor

PR #47, opened 2023-02-02
fix: enforce non zero targets

### LINKED ISSUE #46: governance executor
`createProposal` currently assumes to target `SHORT_EXECUTOR` - which is correct in `99%` of cases.
We should add an overload which allows specifying the executor so it covers `100%`

sakulstra:
- closes https://github.com/bgd-labs/aave-helpers/issues/46

### DIFF
--- a/src/GovHelpers.sol
+++ b/src/GovHelpers.sol
@@ -74,6 +74,22 @@ library GovHelpers {
     internal
     returns (uint256)
   {
+    return _createProposal(AaveGovernanceV2.SHORT_EXECUTOR, delegateCalls, ipfsHash);
+  }
+
+  function createProposal(
+    address executor,
+    Payload[] memory delegateCalls,
+    bytes32 ipfsHash
+  ) internal returns (uint256) {
+    return _createProposal(executor, delegateCalls, ipfsHash);
+  }
+
+  function _createProposal(
+    address executor,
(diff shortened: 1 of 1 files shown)


src/GovHelpers.sol
```diff
@@ -84,6 +100,7 @@ library GovHelpers {
     bytes[] memory calldatas = new bytes[](delegateCalls.length);
     bool[] memory withDelegatecalls = new bool[](delegateCalls.length);
     for (uint256 i = 0; i < delegateCalls.length; i++) {
+      require(delegateCalls[i].target != address(0), 'NON_ZERO_TARGET');
```
sakulstra, line 103: when using this i noticed it's quite easy to messup:
```
GovHelpers.Payload[] memory payloads = new GovHelpers.Payload[](2);
    payloads[0] = GovHelpers.buildMainnet(
      0xd91d1331db4F436DaF47Ec9Dd86deCb8EEF946B4
    );
    vm.startBroadcast();
    GovHelpers.createProposal(
      payloads,
      0x05097b8a0818a75c1db7d54dfd0299581cac0218a058017acb4726f7cc49657e
    );
```

Would not throw, although the `payloads[1]` is empty.
Therefore adding validation.

eboadom (approved):

sendra (approved):

PR #55, opened 2023-02-21
Feat/draft engine 0.2

eboadom:


src/v3-config-engine/V3RateStrategyFactory.sol
```diff
@@ -0,0 +1,139 @@
...
+
+      _strategyByParamsHash[hashedParams] = address(liveStrategies[i]);
+      _strategies.push(address(liveStrategies[i]));
+
+      emit RateStrategyCreated(address(liveStrategies[i]), hashedParams, params);
+    }
+  }
+
+  ///@inheritdoc IV3RateStrategyFactory
+  function createStrategies(RateStrategyParams[] memory params) public returns (address[] memory) {
```
sakulstra, line 41: wondering if there could be some validation on slopes making sense (e.g. ensure stable above variable) - i think for slope 1 is ensured by design, but i think not for slope2.

For excess and ratio, might also make sense to define bounds, as ppl usually don't know what to supply here.

src/v3-config-engine/AaveV3ConfigEngine.sol
```diff
@@ -0,0 +1,520 @@
...
+    ConfiguratorInputTypes.InitReserveInput[]
+      memory initReserveInputs = new ConfiguratorInputTypes.InitReserveInput[](ids.length);
+    address[] memory strategies = RATE_STRATEGIES_FACTORY.createStrategies(rates);
+
+    for (uint256 i = 0; i < ids.length; i++) {
+      uint8 decimals = IERC20Metadata(ids[i]).decimals();
+      require(decimals > 0, 'INVALID_ASSET_DECIMALS');
+
+      initReserveInputs[i] = ConfiguratorInputTypes.InitReserveInput({
+        aTokenImpl: ATOKEN_IMPL,
```
sakulstra: should the engine support custom a/s/v? i'm not 100% sure it should, but might be worth considering an overloaded method which allows supplying them

src/v3-config-engine/V3RateStrategyFactory.sol
eboadom: Yes, agree. Will add something

src/v3-config-engine/AaveV3ConfigEngine.sol
eboadom: I will check the complexity (should be trivial I think), as it could help quite a lot even for the GHO case

eboadom: Done on https://github.com/bgd-labs/aave-helpers/pull/55/commits/814bdaaef88e2af972e6e05220ffa39eb7749c7d

sakulstra (approved):

src/v3-config-engine/V3RateStrategyFactory.sol
eboadom: Will keep this for a following update, because it requires some thinking

src/v3-config-engine/AaveV3ConfigEngine.sol
```diff
@@ -0,0 +1,660 @@
...
+    TokenImplementations implementations;
+  }
+
+  struct Borrow {
+    uint256 enabledToBorrow; // Main config flag, if EngineFlag.DISABLED, some of the other fields will not be considered
+    uint256 flashloanable; // EngineFlag.ENABLED for true, EngineFlag.DISABLED for false otherwise EngineFlag.KEEP_CURRENT
+    uint256 stableRateModeEnabled; // EngineFlag.ENABLED for true, EngineFlag.DISABLED for false otherwise EngineFlag.KEEP_CURRENT
+    uint256 borrowableInIsolation; // EngineFlag.ENABLED for true, EngineFlag.DISABLED for false otherwise EngineFlag.KEEP_CURRENT
+    uint256 withSiloedBorrowing; // EngineFlag.ENABLED for true, EngineFlag.DISABLED for false otherwise EngineFlag.KEEP_CURRENT
+    uint256 reserveFactor; // With 2 digits precision, `10_00` for 10%. Should be positive and < 100_00
```
sakulstra, line 47: should be `<` or `<=`? 3.0.1 should support 100, no?

src/v3-config-engine/AaveV3ConfigEngine.sol
```diff
@@ -0,0 +1,660 @@
...
+      }
+
+      if (borrows[i].withSiloedBorrowing != EngineFlags.KEEP_CURRENT) {
+        POOL_CONFIGURATOR.setSiloedBorrowing(
+          ids[i],
+          EngineFlags.toBool(borrows[i].withSiloedBorrowing)
+        );
+      }
+
+      // TODO: update after v3.0.1
```
sakulstra: ?

src/v3-config-engine/AaveV3ConfigEngine.sol
```diff
@@ -0,0 +1,660 @@
...
+        borrows[i].reserveFactor == EngineFlags.KEEP_CURRENT,
+        'INVALID_RESERVE_FACTOR'
+      );
+
+      if (borrows[i].reserveFactor != EngineFlags.KEEP_CURRENT) {
+        POOL_CONFIGURATOR.setReserveFactor(ids[i], borrows[i].reserveFactor);
+      }
+
+      // TODO: update after v3.0.1
+      if (borrows[i].flashloanable == EngineFlags.ENABLED) {
```
sakulstra, line 320: no update needed after 3.0.1, no?

src/test/AaveV3ConfigEngineGauntletProposal.t.sol
```diff
@@ -0,0 +1,125 @@
...
+import {AaveV3Ethereum} from 'aave-address-book/AaveV3Ethereum.sol';
+import {IAaveV3ConfigEngine} from '../v3-config-engine/IAaveV3ConfigEngine.sol';
+import {AaveV3PolygonRatesUpdates070322} from './mocks/gauntlet-updates/AaveV3PolygonRatesUpdates070322.sol';
+import {AaveV3AvalancheRatesUpdates070322} from './mocks/gauntlet-updates/AaveV3AvalancheRatesUpdates070322.sol';
+import {AaveV3OptimismRatesUpdates070322} from './mocks/gauntlet-updates/AaveV3OptimismRatesUpdates070322.sol';
+import {AaveV3ArbitrumRatesUpdates070322} from './mocks/gauntlet-updates/AaveV3ArbitrumRatesUpdates070322.sol';
+import {DeployEnginePolLib, DeployEngineEthLib, DeployEngineAvaLib, DeployEngineOptLib, DeployEngineArbLib} from '../../script/AaveV3ConfigEngine.s.sol';
+import '../ProtocolV3TestBase.sol';
+
+contract AaveV3PolygonConfigEngineRatesTest is ProtocolV3TestBase {
```
sakulstra: why not using `TestWithExecutor`?

src/test/AaveV3ConfigEngineGauntletProposal.t.sol
```diff
@@ -0,0 +1,125 @@
...
+  using stdStorage for StdStorage;
+
+  function setUp() public {
+    vm.createSelectFork(vm.rpcUrl('polygon'), 40027975);
+  }
+
+  function testEngine() public {
+    IAaveV3ConfigEngine engine = IAaveV3ConfigEngine(
+      DeployEnginePolLib.deploy(0xDd81E6F85358292075B78fc8D5830BE8434aF8BA)
```
sakulstra: imo better put on address-book

src/v3-config-engine/AaveV3ConfigEngine.sol
```diff
@@ -0,0 +1,660 @@
...
+          if (collaterals[i].ltv == EngineFlags.KEEP_CURRENT) {
+            collaterals[i].ltv = currentLtv;
+          }
+
+          if (collaterals[i].liqThreshold == EngineFlags.KEEP_CURRENT) {
+            collaterals[i].liqThreshold = currentLiqThreshold;
+          }
+
+          if (collaterals[i].liqBonus == EngineFlags.KEEP_CURRENT) {
+            collaterals[i].liqBonus = currentLiqBonus - 100_00;
```
sakulstra, line 419: Would add a comment here on why the 100_00 is remove, as it took me quite some time to see it's readded later :sweat_smile:

src/v3-config-engine/AaveV3ConfigEngine.sol
eboadom: On the 3.0.0s 100 will make the protocol malfunction, so to avoid any mistake better to just keep < until we have everything updated. If not we will need to add complexity to review stage, remembering that 100 is wrong

src/v3-config-engine/AaveV3ConfigEngine.sol
brotherlymite: We would need to add a condition when `EngineFlags.ENABLED` for `setReserveFlashLoaning` post `3.0.1` as currently it does not handle that case and would revert otherwise

src/v3-config-engine/AaveV3ConfigEngine.sol
eboadom: This is for the 100% RF, but better to have it more precise

eboadom:
```suggestion
      // TODO: update once all the underlying v3 instances are in 3.0.1 (supporting 100% RF)
```

src/v3-config-engine/AaveV3ConfigEngine.sol
eboadom: Needed, because this allows param change now, not only listing. So somebody could want to disable the flashloanable flag

src/test/AaveV3ConfigEngineGauntletProposal.t.sol
eboadom: Yes, it was waiting until the engine is reviewed, to put both factories and engines

src/v3-config-engine/AaveV3ConfigEngine.sol
eboadom: @brotherlymite

brotherlymite: Sure will add

Makefile
```diff
@@ -6,16 +6,21 @@
...
-listing-engine-pol :;  forge script script/GenericListingEngine.s.sol:DeployListingEnginePol --rpc-url polygon --broadcast --legacy --ledger --mnemonic-indexes ${MNEMONIC_INDEX} --sender ${LEDGER_SENDER} --verify --etherscan-api-key ${ETHERSCAN_API_KEY_POLYGON} -vvvv
-listing-engine-ava :;  forge script script/GenericListingEngine.s.sol:DeployListingEngineAva --rpc-url avalanche --broadcast --legacy --ledger --mnemonic-indexes ${MNEMONIC_INDEX} --sender ${LEDGER_SENDER} --verify --etherscan-api-key ${ETHERSCAN_API_KEY_AVALANCHE} -vvvv
+deploy-engine-opt :;  forge script script/AaveV3ConfigEngine.s.sol:DeployEngineOpt --rpc-url optimism --broadcast --legacy --ledger --mnemonic-indexes ${MNEMONIC_INDEX} --sender ${LEDGER_SENDER} --verify --etherscan-api-key ${ETHERSCAN_API_KEY_OPTIMISM} -vvvv
+deploy-engine-arb :;  forge script script/AaveV3ConfigEngine.s.sol:DeployEngineArb --rpc-url arbitrum --broadcast --legacy --ledger --mnemonic-indexes ${MNEMONIC_INDEX} --sender ${LEDGER_SENDER} --verify --etherscan-api-key ${ETHERSCAN_API_KEY_ARBITRUM} -vvvv
```
kyzia551: No need for a ETHERSCAN_API_KEY_ARBITRUM here. It's already specified in the `foundry.toml`

PR #56, opened 2023-02-22
feat: emit logs

sakulstra:


src/GovHelpers.sol
```diff
@@ -182,13 +196,244 @@ library GovHelpers {
...
+   * @notice This enum contains all possible actions set states
+   */
+  enum ActionsSetState {
+    Queued,
+    Executed,
+    Canceled,
+    Expired
+  }
+
+  // Minimum allowed grace period, which reduces the risk of having an actions set expire due to network congestion
```
eboadom: Any reason to not get rid of this?

src/GovHelpers.sol
```diff
@@ -182,13 +196,244 @@ library GovHelpers {
...
+        actionsSet.values[i],
+        actionsSet.signatures[i],
+        actionsSet.calldatas[i],
+        actionsSet.executionTime,
+        actionsSet.withDelegatecalls[i]
+      );
+      unchecked {
+        ++i;
+      }
+    }
```
eboadom, line 312: Better to keep the `ActionsSetExecuted` event?

sakulstra: idk, i removed all events as they seemed unncessary for me inside the mock, but can add back, no strong opinion

src/GovHelpers.sol
```diff
@@ -182,13 +196,244 @@ library GovHelpers {
...
+
+    ActionsSet storage actionsSet = _actionsSets[actionsSetId];
+    actionsSet.targets = targets;
+    actionsSet.values = values;
+    actionsSet.signatures = signatures;
+    actionsSet.calldatas = calldatas;
+    actionsSet.withDelegatecalls = withDelegatecalls;
+    actionsSet.executionTime = executionTime;
+  }
```
eboadom, line 391: Same as before, no event `ActionsSetQueued` here?

eboadom (approved):

PR #58, opened 2023-03-02
feat: added updateBorrowSide and updatePriceFeeds

brotherlymite:


src/v3-config-engine/IAaveV3ConfigEngine.sol
```diff
@@ -101,7 +101,19 @@ interface IAaveV3ConfigEngine {
 
   /**
    * @dev Example (mock):
-   * CapsUpdate({
+   * PriceFeedUpdate({
+   *   asset: 0x7Fc66500c84A76Ad7e9c93437bFc5Ac33E2DDaE9,
```
eboadom: I think here better to follow the convention an use `AaveV3EthereumAssets.AAVE_UNDERLYING` as example

src/v3-config-engine/IAaveV3ConfigEngine.sol
```diff
@@ -121,6 +133,28 @@ interface IAaveV3ConfigEngine {
     uint256 eModeCategory;
   }
 
+  /**
+   * @dev Example (mock):
```
eboadom, line 137: I would put better as example for example enabling of borrowing, change the RF and the rest, KEEP_CURRENT

src/v3-config-engine/IAaveV3ConfigEngine.sol
brotherlymite: Fixed

src/v3-config-engine/IAaveV3ConfigEngine.sol
brotherlymite: Fixed

src/v3-config-engine/IAaveV3ConfigEngine.sol
```diff
@@ -180,11 +214,25 @@ interface IAaveV3ConfigEngine {
...
   /**
    * @notice Performs an update of the collateral-related params of an asset, in the Aave pool configured in this engine instance
-   * @param updates `CapsUpCollateralUpdatedate[]` list of declarative updates containing the new parameters
+   * @param updates `CollateralUpdate[]` list of declarative updates containing the new parameters
    *   More information on the documentation of the struct.
    */
   function updateCollateralSide(CollateralUpdate[] memory updates) external;
 
+  /**
+   * @notice Performs an update of the price feed of an asset, in the Aave pool configured in this engine instance
```
eboadom: Just realized that these comments (on the other functions too) are not precise, as it is for multiple assets, not only one. Can you change it?

src/v3-config-engine/AaveV3ConfigEngine.sol
```diff
@@ -158,6 +167,15 @@ contract AaveV3ConfigEngine is IAaveV3ConfigEngine {
     _configCollateralSide(configs.ids, configs.collaterals);
   }
 
+  /// @inheritdoc IAaveV3ConfigEngine
+  function updateBorrowSide(BorrowUpdate[] memory updates) public {
+    require(updates.length != 0, 'AT_LEAST_ONE_UPDATE_REQUIRED');
+
+    AssetsConfig memory configs = _repackBorrowUpdate(updates);
+
+    _configBorrowSide(configs.ids, configs.borrows);
```
eboadom, line 177: You will need changes on this function, remember `KEEP_CURRENT` and the types. Always include a mock contract to test the specific update, that way you will notice when compilation fails

eboadom: Generally, it is even better practice to start with how you want the final payload to look, and implement from there up until the engine

src/v3-config-engine/AaveV3ConfigEngine.sol
```diff
@@ -543,6 +561,66 @@ contract AaveV3ConfigEngine is IAaveV3ConfigEngine {
...
+    internal
+    pure
+    returns (AssetsConfig memory)
+  {
+    address[] memory ids = new address[](updates.length);
+    Basic[] memory basics = new Basic[](updates.length);
+
+    for (uint256 i = 0; i < updates.length; i++) {
+      ids[i] = updates[i].asset;
+      basics[i] = Basic({
```
eboadom, line 637: Do you think it is better to keep price-related within Basic? Could be, just wondering

eboadom:
Remember documentation and testing of the new types of updates. Better to always include when opening the PR

src/v3-config-engine/IAaveV3ConfigEngine.sol
brotherlymite: Sure

src/v3-config-engine/AaveV3ConfigEngine.sol
brotherlymite: Ah, my bad. Will add.

src/v3-config-engine/AaveV3ConfigEngine.sol
brotherlymite: We could but in my opinion it's easier to keep it this way as it won't do any updates to other parameters.

src/v3-config-engine/AaveV3ConfigEngine.sol
```diff
@@ -39,11 +39,11 @@ contract AaveV3ConfigEngine is IAaveV3ConfigEngine {
   }
 
   struct Borrow {
-    bool enabledToBorrow; // Main config flag, if false, some of the other fields will not be considered
-    bool flashloanable;
-    bool stableRateModeEnabled;
-    bool borrowableInIsolation;
-    bool withSiloedBorrowing;
+    uint256 enabledToBorrow; // Main config flag, if 0 (false), some of the other fields will not be considered
+    uint256 flashloanable; // 0 for false, 1 for true otherwise KEEP_CURRENT (to be denominated by EngineFlags)
```
eboadom: Here on the documentation, better to indicate EngineFlags.ENABLED and EngineFlags.DISABLED, not referring to 0 and 1

src/v3-config-engine/AaveV3ConfigEngine.sol
```diff
@@ -252,32 +270,38 @@ contract AaveV3ConfigEngine is IAaveV3ConfigEngine {
 
   function _configBorrowSide(address[] memory ids, Borrow[] memory borrows) internal {
     for (uint256 i = 0; i < ids.length; i++) {
-      if (borrows[i].enabledToBorrow) {
+      if (borrows[i].enabledToBorrow == 1 && borrows[i].enabledToBorrow != EngineFlags.KEEP_CURRENT) {
```
eboadom: Better to use abstraction of EngineFlags, not 0 and 1

src/v3-config-engine/AaveV3ConfigEngine.sol
brotherlymite: Understood

src/v3-config-engine/AaveV3ConfigEngine.sol
brotherlymite: Will update

src/v3-config-engine/AaveV3ConfigEngine.sol
```diff
@@ -252,32 +270,38 @@ contract AaveV3ConfigEngine is IAaveV3ConfigEngine {
...
         POOL_CONFIGURATOR.setReserveBorrowing(ids[i], true);
 
         // If enabled to borrow, the reserve factor should always be configured and > 0
         require(
-          borrows[i].reserveFactor > 0 && borrows[i].reserveFactor < 100_00,
+          borrows[i].reserveFactor > 0 && 
+          borrows[i].reserveFactor < 100_00 &&
```
eboadom: For context, on v3.0.1 it is fine to put RF==100%. But better to keep as you have until we upgrade all the v3.0.0 instances.
Could you put a reminder to not forget?

brotherlymite: Sure

src/v3-config-engine/AaveV3ConfigEngine.sol
```diff
@@ -252,32 +270,41 @@ contract AaveV3ConfigEngine is IAaveV3ConfigEngine {
...
           'INVALID_RESERVE_FACTOR'
         );
-        POOL_CONFIGURATOR.setReserveFactor(ids[i], borrows[i].reserveFactor);
 
-        if (borrows[i].stableRateModeEnabled) {
+        if (borrows[i].reserveFactor != EngineFlags.KEEP_CURRENT)
+          POOL_CONFIGURATOR.setReserveFactor(ids[i], borrows[i].reserveFactor);
+
+        if (borrows[i].stableRateModeEnabled == EngineFlags.ENABLED)
           POOL_CONFIGURATOR.setReserveStableRateBorrowing(ids[i], true);
```
eboadom: On conditionals, we use always `{` brackets even if only 1 lines, for both `if` and `if..else`, better to change for consistency

brotherlymite: Will change

src/v3-config-engine/AaveV3ConfigEngine.sol
```diff
@@ -252,32 +270,41 @@ contract AaveV3ConfigEngine is IAaveV3ConfigEngine {
...
+        if (borrows[i].withSiloedBorrowing == EngineFlags.ENABLED)
           POOL_CONFIGURATOR.setSiloedBorrowing(ids[i], true);
-        }
+        else if (borrows[i].withSiloedBorrowing == EngineFlags.DISABLED)
+          POOL_CONFIGURATOR.setSiloedBorrowing(ids[i], false);
+
       }
 
-      if (borrows[i].flashloanable) {
+      // TODO: update after v3.0.1
```
eboadom, line 319: No need for this, because once the underlying pool gets upgrades, it will just work

src/v3-config-engine/AaveV3ConfigEngine.sol
```diff
@@ -252,32 +270,41 @@ contract AaveV3ConfigEngine is IAaveV3ConfigEngine {
...
+          borrows[i].reserveFactor != EngineFlags.KEEP_CURRENT,
           'INVALID_RESERVE_FACTOR'
         );
-        POOL_CONFIGURATOR.setReserveFactor(ids[i], borrows[i].reserveFactor);
 
-        if (borrows[i].stableRateModeEnabled) {
+        if (borrows[i].reserveFactor != EngineFlags.KEEP_CURRENT)
+          POOL_CONFIGURATOR.setReserveFactor(ids[i], borrows[i].reserveFactor);
+
+        if (borrows[i].stableRateModeEnabled == EngineFlags.ENABLED)
```
eboadom: You are missing the KEEP_CURRENT cases on all the following ones, no?

src/v3-config-engine/AaveV3ConfigEngine.sol
brotherlymite: We might have to change for the cases when flashloanable is false

src/v3-config-engine/EngineFlags.sol
```diff
@@ -5,4 +5,10 @@ library EngineFlags {
   /// @dev magic value to be used as flag to keep unchanged any current configuration
   /// Strongly assumes that the value `type(uint256).max - 666` will be used, which seems reasonable
   uint256 public constant KEEP_CURRENT = type(uint256).max - 666;
```
eboadom: Can you change the constant to `42`? 😄

src/v3-config-engine/AaveV3ConfigEngine.sol
brotherlymite: Not so sure, we should not perform any updates when the flag is KEEP_CURRENT right?
Or am I missing something

src/v3-config-engine/EngineFlags.sol
brotherlymite: Sure 😄

src/v3-config-engine/AaveV3ConfigEngine.sol
brotherlymite: Not so sure, if the flag is ENABLED should we check for the KEEP_CURRENT flag

src/v3-config-engine/EngineFlags.sol
```diff
@@ -3,6 +3,22 @@ pragma solidity ^0.8.0;
 
 library EngineFlags {
   /// @dev magic value to be used as flag to keep unchanged any current configuration
-  /// Strongly assumes that the value `type(uint256).max - 666` will be used, which seems reasonable
-  uint256 public constant KEEP_CURRENT = type(uint256).max - 666;
+  /// Strongly assumes that the value `type(uint256).max - 42` will be used, which seems reasonable
```
eboadom:
```suggestion
  /// Strongly assumes that the value `type(uint256).max - 42` will never be used, which seems reasonable
```

PR #68, opened 2023-03-10
feat: rates base payload and factory for v2

brotherlymite:
- Added `AaveV2ConfigEngine` to update rates using factory on aave-v2.
- Added `V2RateStrategyFactory` similar to `V3RateStrategyFactory` but with minor differences in params.
- `DeployV2RatesFactoryLib` deploys the factory and initialises only standard strategies by comparing the contract bytecode length.


src/dependencies/DefaultReserveInterestRateStrategy.sol
```diff
@@ -0,0 +1,257 @@
...
+/**
+ * @title DefaultReserveInterestRateStrategy contract
+ * @notice Implements the calculation of the interest rates depending on the reserve state
+ * @dev The model of interest rate is based on 2 slopes, one before the `OPTIMAL_UTILIZATION_RATE`
+ * point of utilization and another from that one to 100%
+ * - An instance of this same contract, can't be used across different Aave markets, due to the caching
+ *   of the LendingPoolAddressesProvider
+ * @author Aave
+ **/
+contract DefaultReserveInterestRateStrategy is IReserveInterestRateStrategy {
```
eboadom, line 21: Why not importing from the protocol-v2 repository? Versioning 0.8.0? Is there any diff?

script/V2RateStrategyFactory.s.sol
```diff
@@ -0,0 +1,144 @@
...
+
+    address[] memory strategiesOnFactory = ratesFactory.getAllStrategies();
+
+    return (address(ratesFactory), strategiesOnFactory);
+  }
+
+  // To make sure strategies initialised on the factory respect the standard code 
+  function _isStandardStrategy(address strategy) internal view returns (bool) {
+    return (
+      strategy.code.length == 3079 ||
```
eboadom: What are the magic numbers? Important to document from where they come

eboadom: Also, is not better to use some code hashing? And will this work on all v2 instances?

src/dependencies/DefaultReserveInterestRateStrategy.sol
brotherlymite: protocol-v2 repo is still on 0.6.12. Thought to remove the safe math using 0.8.0 but I suppose better to revert it to the older version with safe math and no diff.

script/V2RateStrategyFactory.s.sol
brotherlymite: Will fix and put a comment.

src/dependencies/DefaultReserveInterestRateStrategy.sol
eboadom: I think updating is fine, just we need an explicit diff file

brotherlymite: Sure, will add the diff

script/V2RateStrategyFactory.s.sol
brotherlymite: Fixed

script/V2RateStrategyFactory.s.sol
```diff
@@ -0,0 +1,154 @@
...
+  }
+
+  // To make sure strategies initialised on the factory respect the standard code
+  // We do so by checking if the strategy to initialise matches the standard deployed rates address codehash
+  function _isStandardStrategy(address strategy) internal view returns (bool) {
+    return (
+      strategy.codehash == ETH_STANDARD_RATES_ADDR.codehash || // Current standard codehash of Eth v2 rates strategy
+      strategy.codehash == POL_STANDARD_RATES_ADDR.codehash || // Current standard codehash of Polygon v2 rates strategy
+      strategy.codehash == ETH_AMM_STANDARD_RATES_ADDR.codehash || // Current standard codehash of Eth Amm v2 rates strategy
+      strategy.codehash == AVAX_STANDARD_RATES_ADDR.codehash || // Current standard codehash of Avax v2 rates strategy
```
sakulstra:
```suggestion
```

I think better to align all on the current version. Not much value in allowing to change to an outdated version imo

sakulstra (approved):

src/v2-config-engine/AaveV2ConfigEngine.sol
```diff
@@ -0,0 +1,113 @@
...
+
+import {EngineFlags} from '../v3-config-engine/EngineFlags.sol';
+import './IAaveV2ConfigEngine.sol';
+
+/**
+ * @dev Helper smart contract abstracting the complexity of changing rates configurations on Aave v2.
+ *      It is planned to be used via delegatecall, by any contract having appropriate permissions to update rates
+ * @author BGD Labs
+ */
+contract AaveV2ConfigEngine is IAaveV2ConfigEngine {
```
eboadom, line 13: This needs a comment saying that the contract should be always stateless, with some IMPORTANT or any type of warning

eboadom (approved):

PR #77, opened 2023-04-05
fix: Skip fetch config from non-default IRs

### LINKED ISSUE #76: Fetch of strategy configuration will fail when working with non default IRs
ReserveInterestRateStrategy contracts based on [IReserveInterestRateStrategy](https://github.com/aave/aave-v3-core/blob/master/contracts/interfaces/IReserveInterestRateStrategy.sol) will not work with the current ListingEngine, because will try to fetch data from functions that are part of [IDefaultInterestRateStrategy](https://github.com/aave/aave-v3-core/blob/master/contracts/interfaces/IDefaultInterestRateStrategy.sol).

https://github.com/bgd-labs/aave-helpers/blob/0a774604ebf4019a9dc44768e2085a1ea053d6db/src/ProtocolV3TestBase.sol#L285

miguelmtzinf:
Closes #76

### DIFF
--- a/src/ProtocolV2TestBase.sol
+++ b/src/ProtocolV2TestBase.sol
@@ -49,10 +49,10 @@ contract ProtocolV2TestBase is CommonTestBase {
    * @param pool the pool to be snapshotted
    * @return ReserveConfig[] list of configs
    */
-  function createConfigurationSnapshot(
-    string memory reportName,
-    ILendingPool pool
-  ) public returns (ReserveConfig[] memory) {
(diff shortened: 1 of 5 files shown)


hexonaut:
An alternative is to allow overrides instead of just outputting an error message like this: https://github.com/marsfoundation/aave-helpers/blob/custom-irs/src/ProtocolV3TestBase.sol#L288

Example usage: https://github.com/marsfoundation/sparklend/blob/b41ca787e7d8699fec1ab63c8d122191c6d6a151/src/proposals/20230525/SparkEthereum_20230525.t.sol#L106

PR #88, opened 2023-04-26
Add preliminary support for updating emode category parameters with the config engine

reem:
Still needs some finishing touches like a _validate function for emode category changes and more checks on the params before calling into the pool configurator, and the new engine would need to be deployed before this was usable downstream.


src/v3-config-engine/AaveV3PayloadBase.sol
```diff
@@ -92,6 +94,12 @@ abstract contract AaveV3PayloadBase {
       );
     }
 
+    if (eModeCategories.length != 0) {
```
eboadom, line 97: As `setEModeCategory()` on PoolConfigurator is used both to create a new e-mode category and update an existing one, better to have this step before listings.

That way it is possible to, in the same payload and without boiling down to `preExecute()`, create an e-mode category and assign a new listing to it.
Not something too frequent I would say, but possible and doesn't hurt

src/v3-config-engine/EngineFlags.sol
```diff
@@ -6,6 +6,14 @@ library EngineFlags {
   /// Strongly assumes that the value `type(uint256).max - 42` will never be used, which seems reasonable
   uint256 internal constant KEEP_CURRENT = type(uint256).max - 42;
 
+  /// @dev magic value to be used as flag to keep unchanged any current configuration
+  /// Strongly assumes that the value `AaveV3ConfigEngine.EngineFlags.KEEP_CURRENT_STRING` will never be used, which seems reasonable
+  string internal constant KEEP_CURRENT_STRING = 'AaveV3ConfigEngine.EngineFlags.KEEP_CURRENT_STRING';
```
eboadom, line 11: Maybe not necessary to have so long string on this, as I don't see any chance for `KEEP_CURRENT_STRING` to be used anyhow in the context of any string config on Aave

src/v3-config-engine/EngineFlags.sol
```diff
@@ -6,6 +6,14 @@ library EngineFlags {
...
+
+  /// @dev magic value to be used as flag to keep unchanged any current configuration
+  /// Strongly assumes that the value `0x00000000000000000000000000000000000042` will never be used, which seems reasonable
+  address internal constant KEEP_CURRENT_ADDRESS = address(0x00000000000000000000000000000000000042);
```
eboadom, line 15: I'd say better something different. Projects tend to use the `42`, so even if maybe not clashing now, you never now.
E.g. https://optimistic.etherscan.io/address/0x4200000000000000000000000000000000000000

src/v3-config-engine/AaveV3ConfigEngine.sol
```diff
@@ -453,6 +472,59 @@ contract AaveV3ConfigEngine is IAaveV3ConfigEngine {
     }
   }
 
+  function _configEModeCategories(EModeCategories[] memory updates) internal {
+    for (uint256 i = 0; i < updates.length; i++) {
+        if (
```
eboadom, line 477: A case that is missing here (and actually in some other functions that we are fixing) is all params == KEEP_CURRENT.
Right now it will still query the pool and call `setEModeCategory`, which even if not damaging, it is unnecessary

PR #89, opened 2023-05-15
fix: resolve slight inconsistency

sakulstra:
- there were two tests using the same snapshot name always overwriting each other
- the snapshot only covered `oracle.decimals` but not `oracle.DECIMALS`


kyzia551 (changes requested):
Please move linting into independent PR

kyzia551 (approved):

PR #90, opened 2023-05-17
feat: ipfs tools

sakulstra:
Todo:
- [x] move to gov helpers or separate library
- [x] find a good name for the cli `aave-reports-engine` is to narrow when we eventually use it for lots of things
- [ ] add docs to aave-reports-engine repo (docs are very sparse, but will improve over time) https://github.com/bgd-labs/report-engine


eboadom:
> find a good name for the cli aave-reports-engine is to narrow when we eventually use it for lots of things

Just `aave-cli` ?

sendra (approved):

PR #103, opened 2023-06-05
feat: finalize v2 addresses

sakulstra:


src/v2-config-engine/AaveV2PayloadAvalanche.sol
```diff
@@ -10,7 +10,7 @@ import './AaveV2PayloadBase.sol';
  */
 // TODO: Add rates factory address after deploying
 abstract contract AaveV2PayloadAvalanche is
-  AaveV2PayloadBase(IEngine(address(0)))
+  AaveV2PayloadBase(IEngine(AaveV2Avalanche.LISTING_ENGINE))
 {
   function getPoolContext() public pure override returns (IEngine.PoolContext memory) {
```
sakulstra: should we remove `getPoolContext` ?
Seems unnecessary as there's no listing being supported with v2 engine. I assume also doesn't hurt.

eboadom: Yes, can be removed

PR #108, opened 2023-06-08
feat: use lib to read flags instead of doing calls

sakulstra:


src/test/ProtocolV3TestBase.t.sol
```diff
@@ -56,3 +56,13 @@ contract ProtocolV3TestE2ETestAll is ProtocolV3_0_1TestBase {
...
   }
 }
+
+contract ProtocolV3TestE2ETestSnapshot is ProtocolV3_0_1TestBase {
+  function setUp() public {
+    vm.createSelectFork('optimism', 105213914);
+  }
+
+  function test_snapshot() public {
+    createConfigurationSnapshot('snpashot', AaveV3Optimism.POOL, true, false, false, false);
```
eboadom:
```suggestion
    createConfigurationSnapshot('snapshot', AaveV3Optimism.POOL, true, false, false, false);
```

PR #110, opened 2023-06-14
feat: patch deal (reverting caller)

sakulstra:
THis patched version of deal reads the current caller, deals and reverts to the caller in case it has changed in the meantime

### DIFF
--- a/src/CommonTestBase.sol
+++ b/src/CommonTestBase.sol
@@ -39,43 +39,58 @@ contract CommonTestBase is Test {
    * @param user the user to deal to
    * @param amount the amount to deal
    */
-  function _patchedDeal(address asset, address user, uint256 amount) internal {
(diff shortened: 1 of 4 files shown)


src/CommonTestBase.sol
```diff
@@ -91,10 +107,10 @@ contract CommonTestBase is Test {
       // USDT
       if (asset == AaveV2EthereumAssets.USDT_UNDERLYING) {
         PatchedIERC20(asset).approve(spender, amount);
-        return;
       }
+    } else {
```
eboadom, line 81: Maybe worth it to put the `forceApprove` of safeERC20? https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/utils/SafeERC20.sol#L76

eboadom (approved):

PR #122, opened 2023-07-07
cleanup: align v2 v3

sakulstra:


src/ProtocolV2TestBase.sol
```diff
@@ -163,9 +163,9 @@ contract ProtocolV2TestBase is CommonTestBase {
   }
 
   /**
-   * @dev returns the first collateral in the list that cannot be borrowed in stable mode
+   * @dev returns a "good" collateral in the list that cannot be borrowed in stable mode
```
sakulstra, line 166: it does not actually return the first :sweat_smile:

src/ProtocolV3TestBase.sol
```diff
@@ -195,16 +200,28 @@ contract ProtocolV3TestBase is CommonTestBase {
...
+        // usable as collateral
         configs[i].usageAsCollateralEnabled &&
-        !configs[i].stableBorrowRateEnabled
+        // not stable borrowable as this makes testing stable borrowing unnecessary hard to reason about
+        !configs[i].stableBorrowRateEnabled &&
+        // supply cap not yet reached
+        ((configs[i].supplyCap * 10 ** configs[i].decimals) >
+          IERC20(configs[i].aToken).totalSupply()) &&
+        (// supply cap margin big enough
+        (configs[i].supplyCap * 10 ** configs[i].decimals) -
```
sakulstra, line 222: makes the tests a bit more stable as a collateral is selected that can actually be deposited.

sendra (approved):

PR #123, opened 2023-07-12
Polygon ZkEVM L2 forwarder

0xdapper:


hexonaut:
I like this pattern of using a constructor-supplied immutable for the L2 executor instead of hardcoding the Aave-specific executor as a constant. I will open a PR to update for the previous ones (Optimism, Arbitrum, etc).

hexonaut:
Added here: https://github.com/bgd-labs/aave-helpers/pull/124

PR #128, opened 2023-08-08
Fix aave cli reference

hexonaut:
npm doesn't have `aave-cli`. I've updated the refrence so this tool works.


sakulstra:
superseded by #133

PR #130, opened 2023-08-10
Use collateral with ltv > 15%

hexonaut:
We make an assumption "good collateral" has an ltv of at least 10% in the $1000 collateral vs $100 borrows. Putting 15% for some small buffer.


sakulstra:
Hello,
wondering why does it matter? I assume that with lower lt you will just need more, which is still guaranteed by supplyCaps check, no?

hexonaut:
`e2eTest()` will fail in Spark because the DAI market is first and we've set the LTV/LT to 1bps. DAI market is picked as "good collateral" and there isn't enough LTV power to support borrow $100 worth of other asset vs $1000 worth of DAI.

PR #133, opened 2023-08-11
fea: use npx@

sakulstra:


sakulstra:
This is a more correct version of: https://github.com/bgd-labs/aave-helpers/pull/128

- `--yes` is needed as otherwise on most systems, a prompt would be shown, which foundry cannot handle.
- `-s` is needed as otherwise npm logs install data, which again foundry cannot handle.

Please note though that this has some drawbacks due to the nature of how `forge` and `npx` work.
Forge tries to run test concurrently, which means it will run `npx` concurrently.

`npx` on the other hand isn't quite good at handling concurrency on install so in some test env i managed to break the npx installation with concurrent writes to node modules.
When @bgd-labs/aave-cli is already installed to the project though, non of this matters as npx will take the pre-installed version.

sendra (approved):

PR #137, opened 2023-08-15
Aave Polygon -> Mainnet ERC20 Bridge

efecarranza:
# Changelog

Add bridge to withdraw tokens from Polygon to Mainnet.
Add tests.
Add README.

Once approved, need to deploy on both chains and add address to the README and also to `aave-address-book`.


src/bridges/AavePolEthERC20Bridge.sol
```diff
@@ -0,0 +1,89 @@
...
+
+    IERC20(token).transfer(address(AaveV3Ethereum.COLLECTOR), balance);
+    emit WithdrawToCollector(token, balance);
+  }
+
+  /*
+   * @notice Transfer any tokens accidentally sent to this contract to Aave V3 Polygon Collector
+   * @param tokens List of token addresses
+   */
+  function rescueTokens(address[] calldata tokens) external {
```
sendra: instead of this method. contract should inherit from solidity-utils Rescuable (https://github.com/bgd-labs/solidity-utils/blob/main/src/contracts/utils/Rescuable.sol)

src/bridges/AavePolEthERC20Bridge.sol
```diff
@@ -0,0 +1,89 @@
...
+   * This function withdraws an ERC20 token from Polygon to Mainnet. exit() needs
+   * to be called on mainnet with the corresponding burnProof in order to complete.
+   * @notice Polygon only. Function will revert if called from other network.
+   * @param token Polygon address of ERC20 token to withdraw
+   * @param amount Amount of tokens to withdraw
+   */
+  function withdraw(address token, uint256 amount) external {
+    if (block.chainid != POLYGON_CHAIN_ID) revert InvalidChain();
+
+    IERC20(token).withdraw(amount);
```
sendra: As it looks like this `withdraw` specific method is needed on the ERC20. Will this work only for USDC. Maybe would make sense to check / force that token can be withdrawn?

src/bridges/AavePolEthERC20Bridge.sol
```diff
@@ -0,0 +1,89 @@
...
+   * Withdraws tokens on Mainnet contract to Aave V3 Collector.
+   * @notice Mainnet only. Function will revert if called from other network.
+   * @param token Mainnet address of token to withdraw to Collector
+   */
+  function withdrawToCollector(address token) external {
+    if (block.chainid != ETHEREUM_CHAIN_ID) revert InvalidChain();
+
+    uint256 balance = IERC20(token).balanceOf(address(this));
+
+    IERC20(token).transfer(address(AaveV3Ethereum.COLLECTOR), balance);
```
brotherlymite: I would use `safeTransfer()` instead of `transfer()`.

src/bridges/AavePolEthERC20Bridge.sol
```diff
@@ -0,0 +1,89 @@
...
+  address public constant ROOT_CHAIN_MANAGER = 0xA0c68C638235ee32657e8f720a23ceC1bFc77C77;
+
+  /*
+   * This function withdraws an ERC20 token from Polygon to Mainnet. exit() needs
+   * to be called on mainnet with the corresponding burnProof in order to complete.
+   * @notice Polygon only. Function will revert if called from other network.
+   * @param token Polygon address of ERC20 token to withdraw
+   * @param amount Amount of tokens to withdraw
+   */
+  function withdraw(address token, uint256 amount) external {
```
brotherlymite: @efecarranza Is it a good idea to rename `withdraw()` to something different, maybe `bridge()` or `bridgeToMainnet()` or alternatively rename `withdrawToCollector()` as it gets little confusing, as one is transferring to collector and other is bridging.

src/bridges/AavePolEthERC20Bridge.sol
```diff
@@ -0,0 +1,89 @@
...
+}
+
+contract AavePolEthERC20Bridge is IAavePolEthERC20Bridge {
+  error InvalidChain();
+
+  event Exit();
+  event Withdraw(address token, uint256 amount);
+  event WithdrawToCollector(address token, uint256 amount);
+
+  uint256 public constant ETHEREUM_CHAIN_ID = 1;
```
sakulstra: you can instead rely on https://github.com/bgd-labs/aave-helpers/blob/master/src/ChainIds.sol

src/bridges/DeployBridges.s.sol
```diff
@@ -0,0 +1,19 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import {EthereumScript, PolygonScript} from 'src/ScriptUtils.sol';
+import {AavePolEthERC20Bridge} from './AavePolEthERC20Bridge.sol';
+
+contract DeployEthereum is EthereumScript {
+  function run() external broadcast {
+    bytes32 salt = keccak256(abi.encode(tx.origin, uint256(0)));
```
sendra: provably better to use some string indicating the purpose of the contract as salt

src/bridges/DeployBridges.s.sol
```diff
@@ -0,0 +1,19 @@
...
+    new AavePolEthERC20Bridge{salt: salt}();
```
sendra: Have you double checked, that the deployed address will be the same in all networks, independently of the deployer address?

efecarranza: I tested deploying with my script:

https://polygonscan.com/address/0xCa807a3E47684caee82Fda347729788639Ab9Ee8

https://etherscan.io/address/0xca807a3e47684caee82fda347729788639ab9ee8

Also did the whole flow with these deployed addresses to bridge 1 USDC from Polygon to Mainnet.

src/bridges/DeployBridges.s.sol
efecarranza: Yeah, it was just a placeholder I saw, we can make this anything really for the salt. It's the built-in create2 so it could even be `salt: 0`. If you have any suggestions on the string, happy to update.

src/bridges/AavePolEthERC20Bridge.sol
efecarranza: Good call!

src/bridges/AavePolEthERC20Bridge.sol
efecarranza: @sendra not opposed to this, but who can rescue on Polygon?

src/bridges/AavePolEthERC20Bridge.sol
efecarranza: I don't know if ALL tokens have a withdraw() function, but all the ones I've checked that are supported on AaveV3/AaveV2 do support it.  If called on a token that doesn't have it, it'll revert.

src/bridges/DeployBridges.s.sol
sendra: dont know something like: 'Aave Treasury Bridge' or so

src/bridges/AavePolEthERC20Bridge.sol
sendra: Better to put the Polygon  Guardian for now: 0x1450F2898D6bA2710C98BE9CAF3041330eD5ae58

efecarranza:
@sendra updated!

src/bridges/AavePolEthERC20Bridge.sol
```diff
@@ -0,0 +1,81 @@
...
+    if (block.chainid != ChainIds.MAINNET) revert InvalidChain();
+
+    uint256 balance = IERC20(token).balanceOf(address(this));
+
+    IERC20(token).safeTransfer(address(AaveV3Ethereum.COLLECTOR), balance);
+    emit WithdrawToCollector(token, balance);
+  }
+
+  function whoCanRescue() public pure override returns (address) {
+    return AaveV2Polygon.EMERGENCY_ADMIN;
```
brotherlymite: @efecarranza hardcoding the polygon's guardian address as the recuable address could cause some issues when we call `rescueTokens()` on ethereum, it would rescue to the address of guardian of polygon on mainnet, and funds will be stuck.
I would rather make the contract ownable and then keep `whoCanRescue()` as the executor contract. This way we can change `whoCanRescue()` later on as well while we migrate from gov v2 -> gov v3.

efecarranza: hey @brotherlymite the ethereum guardian was suggested by pol above so just went with that, but you're right, it would cause problems as there's no chain check, i'm not opposed to making it ownable but i  also, i liked the previous `rescueTokens` method because it was checking chains while `withdrawToCollector` is the mainnet.

i see two paths forward:

1. make ownable
2. don't extend rescuable and keep the `rescueTokens` method as is.

what do you guys think? @sendra

sendra: I think should be option 1. As we are working with token balances, multiple problems could happen that would make the tokens to be stuck on this contract. By having rescuable + ownable, we ensure that they can be saved in any situation (if the collector would get compromised for example)

efecarranza:
@brotherlymite @sendra 

i made it ownable, updated tests. now that it's ownable, should we make `bridge()` onlyOwner? probably, right?

sendra:
> i made it ownable, updated tests. now that it's ownable, should we make bridge() onlyOwner? probably, right?

Yes i think it would make sense. Also don't forget to add on the script the transfer of ownership

sendra (approved):

PR #138, opened 2023-08-16
Aave COW Swaps

efecarranza:
# Changelog

Add AaveCOWSwap contract.
Add helper payloads.
Add tests.
Add README.

### DIFF
--- a/src/swaps/AaveCOWSwaps.sol
+++ b/src/swaps/AaveCOWSwaps.sol
@@ -1,0 +1,176 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity 0.8.19;
+
+import {IERC20} from 'solidity-utils/contracts/oz-common/interfaces/IERC20.sol';
+import {SafeERC20} from 'solidity-utils/contracts/oz-common/SafeERC20.sol';
+import {OwnableWithGuardian} from 'solidity-utils/contracts/access-control/OwnableWithGuardian.sol';
+import {Initializable} from 'solidity-utils/contracts/transparent-proxy/Initializable.sol';
+import {AaveV3Ethereum} from 'aave-address-book/AaveV3Ethereum.sol';
+
+import {IPriceChecker} from './interfaces/IExpectedOutCalculator.sol';
+import {IMilkman} from './interfaces/IMilkman.sol';
+
+contract AaveCOWSwaps is Initializable, OwnableWithGuardian {
+  using SafeERC20 for IERC20;
+
+  event SwapCanceled(address fromToken, address toToken, uint256 amount);
+  event SwapRequested(address fromToken, address toToken, uint256 amount);
+  event TokenUpdated(address indexed token, bool allowed);
+
+  error Invalid0xAddress();
+  error InvalidAmount();
+  error InvalidRecipient();
+  error InvalidToken();
+  error OracleNotSet();
+
+  address public constant BAL80WETH20 = 0x5c6Ee304399DBdB9C8Ef030aB642B10820DB8F56;
+  address public constant BPT_PRICE_CHECKER = 0xBeA6AAC5bDCe0206A9f909d80a467C93A7D6Da7c;
+
+  address public chainlinkPriceChecker = 0xe80a1C615F75AFF7Ed8F08c9F21f9d00982D666c;
+  address public milkman = 0x11C76AD590ABDFFCD980afEC9ad951B160F02797;
+
+  /// @notice Chainlink Oracle address for given token (supports only USD bases)
+  mapping(address tokenAddress => address) public tokenChainlinkOracle;
+
+  function initialize() external initializer {
+    _transferOwnership(_msgSender());
+    _updateGuardian(_msgSender());
+  }
+
+  function swap(
+    address fromToken,
+    address toToken,
+    address recipient,
+    uint256 amount,
+    uint256 slippage
+  ) external onlyOwner {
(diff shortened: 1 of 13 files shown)


src/swaps/AaveCOWSwaps.sol
```diff
@@ -0,0 +1,170 @@
...
+
+  error Invalid0xAddress();
+  error InvalidAmount();
+  error InvalidRecipient();
+  error OracleNotSet();
+
+  address public constant BAL80WETH20 = 0x5c6Ee304399DBdB9C8Ef030aB642B10820DB8F56;
+  address public constant BPT_PRICE_CHECKER = 0xBeA6AAC5bDCe0206A9f909d80a467C93A7D6Da7c;
+
+  address public chainlinkPriceChecker = 0xe80a1C615F75AFF7Ed8F08c9F21f9d00982D666c;
```
sendra: We think that it would be better to pass the priceChecker and milkman addresses as input to the swap method, this way you would be forcing the user, to check / get up to date addresses for the specific swap.
This would be safer, as if not, possibility that user just forgets to set the new addresses could be high. It would also make it so you don't need to hardcode BAL token and price checker.

Then a section on README could be added on how to find up to date addresses, and / or an other option could be to add these kind of addresses in a different lib (maybe like address-book) depending on how often they get updated

src/swaps/AaveCOWSwaps.sol
```diff
@@ -0,0 +1,170 @@
...
+    IMilkman(milkman).requestSwapExactTokensForTokens(
+      amount,
+      IERC20(fromToken),
+      IERC20(toToken),
+      recipient,
+      priceChecker,
+      data
+    );
+
+    emit SwapRequested(fromToken, toToken, fromOracle, toOracle, amount, recipient);
```
sendra: would need to emit milkman address also right? as its needed in the cancel swap input, and could change between swap and cancel?
Also the slippage?

efecarranza: Unfortunately there's no way to get the milkman address, it's only available via event so we'd have to retrieve it to get it. Added slippage.

efecarranza:
@sendra updated!

src/swaps/AaveSwapper.sol
```diff
@@ -12,29 +12,35 @@ import {AaveV3Ethereum} from 'aave-address-book/AaveV3Ethereum.sol';
...
   event SwapCanceled(address fromToken, address toToken, uint256 amount);
-  event SwapRequested(address fromToken, address toToken, address fromOracle, address toOracle, uint256 amount, address recipient);
+  event SwapRequested(
+    address fromToken,
+    address toToken,
+    address fromOracle,
+    address toOracle,
+    uint256 amount,
+    address recipient,
+    uint256 slippage
```
sendra, line 27: I would also add milkman address (passed in params) and also indexing from/to token and recipient

src/swaps/AaveSwapper.sol
```diff
@@ -12,29 +12,35 @@ import {AaveV3Ethereum} from 'aave-address-book/AaveV3Ethereum.sol';
 import {IPriceChecker} from './interfaces/IExpectedOutCalculator.sol';
 import {IMilkman} from './interfaces/IMilkman.sol';
 
-contract AaveCOWSwaps is Initializable, OwnableWithGuardian, Rescuable {
+contract AaveSwapper is Initializable, OwnableWithGuardian, Rescuable {
   using SafeERC20 for IERC20;
 
   event SwapCanceled(address fromToken, address toToken, uint256 amount);
```
sendra: would index from / to token

efecarranza:
@sendra updated.

src/swaps/README.md
```diff
@@ -0,0 +1,131 @@
...
+
+AaveCOWSwaps is a permissioned smart contract, and it has two potential privileged users: the owner and the guardian.
+The owner will be the DAO (however, ownership can be transferred) and the guardian is an address to be chosen by the
+DAO to more easily cancel swaps without relying on governance. AaveCOWSwaps can hold funds and swap for other tokens
+and keep them in case the DAO chooses so. AaveCOWSwaps can only withdraw tokens to the Collector contract.
+
+### Methods
+
+```
+function swap(
```
sendra, line 25: Should be updated to the new interface

src/swaps/README.md
```diff
@@ -0,0 +1,131 @@
...
+## How It Works
+
+AaveCOWSwaps relies on [Milkman](https://github.com/charlesndalton/milkman), a smart contract that builds on top of
+[COW Swap](https://swap.cow.fi/#/faq/protocol), under the hood in order to find the best possible swap execution for
+the DAO while protecting funds from MEV exploits and bad slippage.
+
+AaveCOWSwaps is a permissioned smart contract, and it has two potential privileged users: the owner and the guardian.
+The owner will be the DAO (however, ownership can be transferred) and the guardian is an address to be chosen by the
+DAO to more easily cancel swaps without relying on governance. AaveCOWSwaps can hold funds and swap for other tokens
+and keep them in case the DAO chooses so. AaveCOWSwaps can only withdraw tokens to the Collector contract.
```
brotherlymite: The AaveCOWSwaps contract can withdraw to any address right as it is inheriting from `Rescuable`

sendra: yes

efecarranza:
@sendra updated docs.

PR #166, opened 2023-11-03
PolEthBridge:  Update bridge owner and README as well as add missing natspec

efecarranza:
# Changelog

Update deployment script with correct owner.
Add missing natspec of Bridge contract.
Update README with information on how to process multiple burns in a single transaction.


sendra:
I think, now that its being updated. It would be better to move events, natspec, and so on to the interface, and then use @inheritdocs.

efecarranza:
@sendra updated!

src/bridges/AavePolEthERC20Bridge.sol
```diff
@@ -34,7 +34,7 @@ contract AavePolEthERC20Bridge is Ownable, Rescuable, IAavePolEthERC20Bridge {
 
   error InvalidChain();
 
-  event Exit();
+  event Exit(bytes proof);
   event Bridge(address token, uint256 amount);
   event WithdrawToCollector(address token, uint256 amount);
```
sendra, line 40: Missing natspec on interface

src/bridges/AavePolEthERC20Bridge.sol
```diff
@@ -44,38 +44,34 @@ contract AavePolEthERC20Bridge is Ownable, Rescuable, IAavePolEthERC20Bridge {
     _transferOwnership(_owner);
```
sendra, line 48: constructor missing natspec (on this file)

sendra:
address public constant ROOT_CHAIN_MANAGER missing natspec on interface (as this code was not change, could not leave comment directly on code)

efecarranza:
@sendra updated, thank you. i added a new function and update the README accordingly to add a convenience function to check whether a token is mapped prior to burning.

Zer0dot (approved):

efecarranza:
@Zer0dot alright, i've added the relevant TXs and now it forward ETH as well (when withdrawing WETH)

sendra (approved):

Zer0dot (approved):

PR #185, opened 2023-11-24
AaveSwapper: Add test scripts

efecarranza:
# Changelog

Update README with instructions to perform live test.
Add scripts to easily test with owned set up.


src/swaps/scripts/TestSwap.s.sol
```diff
@@ -0,0 +1,69 @@
...
+
+pragma solidity 0.8.18;
+
+import {IERC20} from 'solidity-utils/contracts/oz-common/interfaces/IERC20.sol';
+
+import {EthereumScript} from 'src/ScriptUtils.sol';
+import {AaveSwapper} from 'src/swaps/AaveSwapper.sol';
+
+// make test-swap
+contract TestSwap is EthereumScript {
```
brotherlymite: I would put the script in the `./scripts` directory instead of `./src/swaps/scripts`

src/swaps/scripts/TestSwap.s.sol
```diff
@@ -0,0 +1,69 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity 0.8.18;
```
brotherlymite:
```suggestion
pragma solidity ^0.8.0;
```

brotherlymite:
Could you also update the libraries to the latest version (aave-address-book) or update with the latest `master` branch

efecarranza:
@brotherlymite updated!

src/swaps/AaveSwapper.sol
```diff
@@ -41,8 +41,8 @@ contract AaveSwapper is Initializable, OwnableWithGuardian, Rescuable {
   address public constant BAL80WETH20 = 0x5c6Ee304399DBdB9C8Ef030aB642B10820DB8F56;
 
   function initialize() external initializer {
-    _transferOwnership(AaveGovernanceV2.SHORT_EXECUTOR);
-    _updateGuardian(0xA519a7cE7B24333055781133B13532AEabfAC81b);
+    _transferOwnership(0x3765A685a401622C060E5D700D9ad89413363a91);
```
sendra, line 44: what are these addresses? Should this be taken from address book? Provably should be passed as params though

PR #195, opened 2023-12-11
fix: remove cap limitation

sakulstra:
replaces https://github.com/bgd-labs/aave-helpers/pull/194

### DIFF
--- a/src/ProtocolV3TestBase.sol
+++ b/src/ProtocolV3TestBase.sol
@@ -180,7 +180,7 @@ contract ProtocolV3TestBase is CommonTestBase {
    */
   function e2eTest(IPool pool) public {
     ReserveConfig[] memory configs = _getReservesConfigs(pool);
-    ReserveConfig memory collateralConfig = _getGoodCollateral(pool, configs);
+    ReserveConfig memory collateralConfig = _getGoodCollateral(configs);
     uint256 snapshot = vm.snapshot();
     for (uint256 i; i < configs.length; i++) {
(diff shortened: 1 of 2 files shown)


src/ProtocolV3TestBase.sol
```diff
@@ -306,8 +305,7 @@ contract ProtocolV3TestBase is CommonTestBase {
    */
   function _getGoodCollateral(
     IPool pool,
-    ReserveConfig[] memory configs,
-    uint256 minSupplyCapDollarMargin
+    ReserveConfig[] memory configs
   ) private view returns (ReserveConfig memory config) {
```
Zer0dot:
```suggestion
  ) private pure returns (ReserveConfig memory config) {
```

src/ProtocolV3TestBase.sol
```diff
@@ -306,8 +305,7 @@ contract ProtocolV3TestBase is CommonTestBase {
    */
   function _getGoodCollateral(
     IPool pool,
```
Zer0dot:
```suggestion
    IPool,
```

Zer0dot (approved):

src/ProtocolV3TestBase.sol
sakulstra: Ah in that case probably better to just remove completely

Zer0dot: Sure, either works

Zer0dot (approved):

PR #209, opened 2023-12-31
Arbitrum -> Mainnet ERC20 Bridge

efecarranza:
# Changelog

Add contract to bridge tokens to mainnet.
Add tests for contract.
Add README.

Relevant aave-cli PR: https://github.com/bgd-labs/aave-cli/pull/81


src/bridges/arbitrum/AaveArbEthERC20Bridge.sol
```diff
@@ -0,0 +1,105 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity 0.8.19;
```
brotherlymite:
```suggestion
pragma solidity ^0.8.0;
```

Fixed solc versions are a bit difficult to maintain as foundry requires other contracts to also compile with the specific version

foundry.toml
```diff
@@ -8,7 +8,7 @@ remappings = [
 ]
 fs_permissions = [{access = "read-write", path = "./reports"}]
 ffi = true
-solc = '0.8.18'
+solc = '0.8.19'
```
brotherlymite: Any specific reasons to update, if not I think we can keep it this way, as it would cause some inconsistencies with the deployed contracts.

brotherlymite:
@efecarranza could you also merge the latest `master` and resolve the merge conflicts

src/bridges/arbitrum/AaveArbEthERC20Bridge.sol
```diff
@@ -0,0 +1,105 @@
...
+    uint256 amount,
+    bytes calldata data
+  ) external;
+}
+
+contract AaveArbEthERC20Bridge is Ownable, Rescuable, IAaveArbEthERC20Bridge {
+  using SafeERC20 for IERC20;
+
+  /// @notice This function is not supported on this chain
+  error InvalidChain();
```
sendra: Errors should be defined on interface

src/bridges/arbitrum/AaveArbEthERC20Bridge.sol
```diff
@@ -0,0 +1,105 @@
...
+
+  /// @notice Emitted when bridging a token from Arbitrum to Mainnet
+  event Bridge(address token, uint256 amount);
```
sendra: Events should be defined on interface

src/bridges/arbitrum/AaveArbEthERC20Bridge.sol
```diff
@@ -0,0 +1,105 @@
...
+  /// @param _owner The owner of the contract upon deployment
+  constructor(address _owner) {
+    _transferOwnership(_owner);
+  }
+
+  /// @inheritdoc IAaveArbEthERC20Bridge
+  function bridge(
+    address token,
+    address l1Token,
+    address gateway,
```
sendra, line 71: Should gateway be an immutable? Or does the gateway change often?

src/bridges/arbitrum/AaveArbEthERC20Bridge.sol
sendra: Token param provably makes sense to be indexed

src/bridges/arbitrum/AaveArbEthERC20Bridge.sol
```diff
@@ -0,0 +1,105 @@
...
+      l2sender,
+      to,
+      l2block,
+      l1block,
+      l2timestamp,
+      value,
+      data
+    );
+
+    emit Exit();
```
sendra: event should have information of the exit params: token, destination, value, etc...

src/bridges/arbitrum/AaveArbEthERC20Bridge.sol
```diff
@@ -0,0 +1,105 @@
...
+
+interface IL2Gateway {
+  function outboundTransfer(
+    address tokenAddress,
+    address recipient,
+    uint256 amount,
+    bytes calldata data
+  ) external;
+}
+
```
sendra, line 53: Missing Contract natspec (author, notice, etc...)

src/bridges/arbitrum/AaveArbEthERC20Bridge.sol
```diff
@@ -0,0 +1,105 @@
...
+import {IERC20} from 'solidity-utils/contracts/oz-common/interfaces/IERC20.sol';
+import {SafeERC20} from 'solidity-utils/contracts/oz-common/SafeERC20.sol';
+import {Ownable} from 'solidity-utils/contracts/oz-common/Ownable.sol';
+import {Rescuable} from 'solidity-utils/contracts/utils/Rescuable.sol';
+import {AaveV3Ethereum} from 'aave-address-book/AaveV3Ethereum.sol';
+
+import {ChainIds} from '../../ChainIds.sol';
+import {IAaveArbEthERC20Bridge} from './IAaveArbEthERC20Bridge.sol';
+
+interface IL1Outbox {
```
sendra, line 15: Missing natspec of these interfaces

src/bridges/arbitrum/AaveArbEthERC20Bridge.sol
```diff
@@ -0,0 +1,105 @@
...
+
+    emit Bridge(token, amount);
+  }
+
+  /// @inheritdoc IAaveArbEthERC20Bridge
+  function exit(
+    bytes32[] calldata proof,
+    uint256 index,
+    address l2sender,
+    address to,
```
sendra: To should be immutable (treasury) if this method is callable by anyone.

src/bridges/arbitrum/AaveArbEthERC20Bridge.sol
efecarranza: different tokens have different gateways (ie DAI vs USDC)

src/bridges/arbitrum/AaveArbEthERC20Bridge.sol
efecarranza: `to` is the L1 receiving Gateway. it's not who received the tokens. the actual `to` is in the `bridge()` function hardcoded as the Collector :)

src/bridges/arbitrum/AaveArbEthERC20Bridge.sol
efecarranza: unfortunately none of that data is available 😬 when we exit. we could add the sender and the blocks maybe?

src/bridges/arbitrum/AaveArbEthERC20Bridge.sol
efecarranza: done

src/bridges/arbitrum/AaveArbEthERC20Bridge.sol
efecarranza: added

efecarranza:
@brotherlymite @sendra I've updated the PR accordingly and left comments where applicable. thanks!

src/bridges/arbitrum/AaveArbEthERC20Bridge.sol
sendra: yup, would add l2sender, msg.sender, the blocks, to, value, data, etc... basically everything that might be needed for an offchain tracking alerts or interface

src/bridges/arbitrum/AaveArbEthERC20Bridge.sol
```diff
@@ -0,0 +1,115 @@
...
+
+  address public constant MAINNET_OUTBOX = 0x0B9857ae2D4A3DBe74ffE1d7DF045bb7F96E4840;
+
+  /// @param _owner The owner of the contract upon deployment
+  constructor(address _owner) {
+    _transferOwnership(_owner);
+  }
+
+  /// @inheritdoc IAaveArbEthERC20Bridge
+  function bridge(
```
brotherlymite, line 68: @efecarranza what will happen if at the time of bridging the arbitrum sequencer goes down, do you think it make sense to avoid bridging assets when the sequencer is down

brotherlymite, line 68: No strong opinion, but instead of transferring tokens to the bridge and calling the `bridge()` method, does it make sense to do a pull instead of a push. i.e we approve the token amount to the bridge contract and the bridge contract does a `transferFrom()`

src/bridges/arbitrum/AaveArbEthERC20Bridge.sol
brotherlymite: @efecarranza the natspec comment and the variable name is a bit confusing, shall we rename it so people do not get confused.

src/bridges/arbitrum/AaveArbEthERC20Bridge.sol
```diff
@@ -0,0 +1,115 @@
...
+    bytes calldata data
+  ) external;
+}
+
+/// @author efecarranza.eth
+/// @notice Contract to bridge ERC20 tokens from Arbitrum to Mainnet
+contract AaveArbEthERC20Bridge is Ownable, Rescuable, IAaveArbEthERC20Bridge {
+  using SafeERC20 for IERC20;
+
+  address public constant MAINNET_OUTBOX = 0x0B9857ae2D4A3DBe74ffE1d7DF045bb7F96E4840;
```
brotherlymite, line 60:
```suggestion
  /// @inheritdoc IAaveArbEthERC20Bridge
  address public constant MAINNET_OUTBOX = 0x0B9857ae2D4A3DBe74ffE1d7DF045bb7F96E4840;
```

src/bridges/arbitrum/AaveArbEthERC20Bridge.sol
```diff
@@ -0,0 +1,115 @@
+// SPDX-License-Identifier: MIT
+
```
brotherlymite, line 2:
```suggestion
```

src/bridges/arbitrum/AaveArbEthERC20Bridge.sol
efecarranza: hey @brotherlymite asking the Arbitrum team. is there a way to check the sequencer on-chain? thank you

brotherlymite: Yes, we can check it on-chain from the [chainlink sequencer uptime feed](https://docs.chain.link/data-feeds/l2-sequencer-feeds/).

efecarranza: we could check that or alternatively since we can bypass the sequencer it might not be a problem: https://docs.arbitrum.io/sequencer#unhappyuncommon-case-sequencer-isnt-doing-its-job

sendra: I think provably makes sense to leave as is. as by the looks of it, its the gateway that will do the transferFrom, from this contract, so would be confusing, provably, to approve the gateway somewhere else

efecarranza: 👍

src/bridges/arbitrum/AaveArbEthERC20Bridge.sol
efecarranza: @brotherlymite what would you say is confusing so i can update it? this is how it's in the arbitrum docs.

brotherlymite: @efecarranza I meant the `to` param, on natspec. The `to` param is defined as: The address receiving the bridged fund, but instead it is the L1 receiving gateway contract (destination address for L1 contract call).

efecarranza:
@brotherlymite updated!

brotherlymite (approved):

scripts/DeployBridges.s.sol
```diff
@@ -40,9 +43,120 @@ contract DeployPlasmaPolygon is PolygonScript {
...
+      210534980,
+      19854800,
+      1715527345,
+      1000,
+      data
+    );
+  }
+}
+  
+contract BridgeTokens is ArbitrumScript {
```
sendra: if this gets modified every time, depending on the tokens that we need to bridge. then maybe makes more sense to put into other script contract, or just on readme. As doesnt make sense to keep changing this repo every time

scripts/DeployBridges.s.sol
```diff
@@ -40,9 +43,120 @@ contract DeployPlasmaPolygon is PolygonScript {
...
+}
+
+contract DeployArbBridgeArbitrum is ArbitrumScript {
   function run() external broadcast {
     bytes32 salt = 'Aave Treasury Bridge';
     new AaveArbEthERC20Bridge{salt: salt}(0x3765A685a401622C060E5D700D9ad89413363a91);
   }
 }
+
+contract ExitTokens is EthereumScript {
```
sendra: This could also provably be moved to a readme, if this will change with every execution?

sendra (approved):

PR #201, opened 2023-12-20
Aave Swapper: Limit Orders

efecarranza:
# Changelog

Add method to do a limit swap order.
Add method to cancel a limit swap order.
Refactor code to DRY swap functionality.
Update README to add information on limit orders.
Add natspec.

### Tests

Sample ERC20 to ERC20 swap with 18 decimals:  https://explorer.cow.fi/address/0x75a37623F8308eDaA2C21D8B51B91393A9E4033a
https://etherscan.io/tx/0x2186f0cabbbf1930b436783eca69328bfa5a7453c1c7d9a6ed9d7818b3ad9729

Sample ERC20 to ERC20 swap with 6 decimals:
https://explorer.cow.fi/address/0x141764c308997ec2a3AFE2cBAA1307610b126c01
https://etherscan.io/tx/0x135a6218866f050634bf52e699b93a9374c12cba19d2de104aa8296e86d0ce51

PR #225, opened 2024-01-20
Aave Swapper - TWAP Orders

efecarranza:
# Changelog

Add TWAP orders to the Aave Swapper contract.
Add tests for TWAP orders.
Add documentation to README.

Builds on top of: https://github.com/bgd-labs/aave-helpers/pull/201


src/swaps/AaveSwapper.sol
```diff
@@ -85,6 +127,92 @@ contract AaveSwapper is Initializable, OwnableWithGuardian, Rescuable {
...
+  /// @param recipient Address that will receive toToken
+  /// @param sellAmount The amount of tokens to sell per TWAP swap
+  /// @param minPartLimit Minimum amount of toToken to receive per TWAP swap
+  /// @param startTime Timestamp of when TWAP orders start
+  /// @param numParts Number of TWAP swaps to take place (each for sellAmount)
+  /// @param partDuration How long each TWAP takes (ie: hourly, weekly, etc)
+  /// @param span The timeframe the orders can take place in
+  function twapSwap(
+    address handler,
+    address relayer,
```
brotherlymite, line 115: @efecarranza does it make sense to make these params as immutable?

efecarranza: hey @brotherlymite when building the original AaveSwapper, feedback was provided that we should be allowed to pass the specific price checker, so that's why I went with that route for this even though they are not a price checker, but kind of the same concept. We could have it as a contract variable that can be updated as well in case of changes to the cow swap relayer or handler, though I doubt those values are likely to change.

@eboado If my memory serves me well, I think you and I had discussed this approach. Wanted to see if you had any thoughts around it. Thank you

.gitmodules
pavelvm5, line 1: Could you update `composable-cow` lib? Cause it has `balancer` lib inside, which can't be installed right now.

At least I don't have access to it. But the more fresh commit doesn't have such lib in it.

scripts/AaveSwapperDeployment.s.sol
```diff
@@ -12,8 +12,8 @@ contract DeployAaveSwapper is Script {
   function run() external {
     vm.startBroadcast();
 
-    address aaveSwapper = address(new AaveSwapper());
-    ITransparentProxyFactory(MiscEthereum.TRANSPARENT_PROXY_FACTORY).create(
+    address aaveSwapper = address(new AaveSwapper(0xfdaFc9d1902f4e0b84f65F49f244b32b31013b74));
+    TransparentProxyFactory(MiscEthereum.TRANSPARENT_PROXY_FACTORY).create(
```
pavelvm5, line 16:
```suggestion
    ITransparentProxyFactory(MiscEthereum.TRANSPARENT_PROXY_FACTORY).create(
```

src/swaps/AaveSwapper.sol
```diff
@@ -133,33 +269,101 @@ contract AaveSwapper is IAaveSwapper, Initializable, OwnableWithGuardian, Rescua
...
+    IERC20(fromToken).forceApprove(milkman, amount);
+
+    IMilkman(milkman).requestSwapExactTokensForTokens(
+      amount,
+      IERC20(fromToken),
+      IERC20(toToken),
+      recipient,
+      priceChecker,
+      priceCheckerData
+    );
```
pavelvm5, line 302: Function in interface has 7 arguments, not 6. `appData` was skipped.

Сan you make sure that this code compiles and all your tests pass by installing everything from scratch?

pavelvm5 (changes requested):

PR #237, opened 2024-01-31
Some workarounds for GovV3 templating

kyzia551:


src/GovV3PlaygroundAdvanced.sol
```diff
@@ -0,0 +1,147 @@
...
+
+abstract contract DeployPayloads is WithPayloads, Script {
+  function run() external {
+    ActionsPerChain[] memory actionsPerChain = getActions();
+
+    for (uint256 i = 0; i < actionsPerChain.length; i++) {
+      ActionsPerChain memory rawActions = actionsPerChain[i];
+      require(rawActions.actionCode.length != 0, 'should be at least one payload action per chain');
+
+      vm.rpcUrl(rawActions.chainName);
```
sendra, line 29: Here would go the conditional to check that payload is of the appropriate chain right (as we would not be using --multi)? and i guess the contract would still inherit from the base chain payload (like currently does)

src/GovV3PlaygroundAdvanced.sol
```diff
@@ -0,0 +1,119 @@
...
+      _supportedChain[keccak256(bytes(supportedChains[i]))] = true;
+    }
+  }
+
+  function isChainSupported(string memory chain) public view override returns (bool) {
+    return _supportedChain[keccak256(bytes(chain))];
+  }
+}
+
+abstract contract DeployPayloadsSingleChain is DeployPayloads {
```
kyzia551, line 67: @sendra something like this to stay with one chain deployment

PR #261, opened 2024-03-06
Plasma Bridge

efecarranza:
# Changelog

Add Plasma bridge for native MATIC.
Add tests.


src/bridges/polygon/AavePolEthPlasmaBridge.sol
```diff
@@ -0,0 +1,126 @@
...
+/// @author efecarranza.eth
+/// @notice Helper contract to bridge assets from Polygon to Ethereum
+contract AavePolEthPlasmaBridge is Ownable, Rescuable, IAavePolEthPlasmaBridge {
+  using SafeERC20 for IERC20;
+
+  /// @dev The called method is not available on this chain
+  error InvalidChain();
+
+    /// @dev Emitted when a bridge is initiated
+  event Bridge(address token, uint256 amount);
```
sendra: events should go to the interface

src/bridges/polygon/AavePolEthPlasmaBridge.sol
```diff
@@ -0,0 +1,126 @@
...
+  /// @dev Emitted when a token bridge is finalized
+  event Exit(address indexed token);
+
+    /// @dev Emitted when multiple token bridges are finalized
+  event ExitBatch(address[] indexed tokens);
+
+  /// @dev Emitted when token is withdrawn to the Aave Collector
+  event WithdrawToCollector(address token, uint256 amount);
+
+  /// @dev The mainnet address of the Predicate contract to confirm withdrawal
```
sendra: natspec on constants should go to interface

src/bridges/polygon/IAavePolEthPlasmaBridge.sol
```diff
@@ -0,0 +1,28 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.0;
+
+interface IAavePolEthPlasmaBridge {
```
sendra, line 8: Add natspec

scripts/DeployBridges.s.sol
```diff
@@ -19,3 +21,25 @@ contract DeployPolygon is PolygonScript {
     new AavePolEthERC20Bridge{salt: salt}(GovernanceV3Polygon.EXECUTOR_LVL_1);
   }
 }
+
+contract DeployPlasmaEthereum is EthereumScript {
+  function run() external broadcast {
+    bytes32 salt = 'Aave Treasury Plasma Bridge';
+    new AavePolEthPlasmaBridge{salt: salt}(0x3765A685a401622C060E5D700D9ad89413363a91);
```
sendra, line 28: I guess, after this is tested in production, ownership will change to Executor?

src/bridges/polygon/AavePolEthPlasmaBridge.sol
```diff
@@ -0,0 +1,126 @@
...
+
+    IERC20Polygon(token).withdraw{value: amount}(amount);
+    emit Bridge(token, amount);
+  }
+
+  /// @inheritdoc IAavePolEthPlasmaBridge
+  function confirmExit(bytes calldata burnProof) external {
+    if (block.chainid != ChainIds.MAINNET) revert InvalidChain();
+
+    IERC20PredicateBurnOnly(0x158d5fa3Ef8e4dDA8a5367deCF76b94E7efFCe95).startExitWithBurntTokens(
```
sendra:
```suggestion
    IERC20PredicateBurnOnly(ERC20_PREDICATE_BURN).startExitWithBurntTokens(
```

src/bridges/polygon/AavePolEthPlasmaBridge.sol
```diff
@@ -0,0 +1,126 @@
...
+  event ExitBatch(address[] indexed tokens);
+
+  /// @dev Emitted when token is withdrawn to the Aave Collector
+  event WithdrawToCollector(address token, uint256 amount);
+
+  /// @dev The mainnet address of the Predicate contract to confirm withdrawal
+  address public constant ERC20_PREDICATE_BURN = 0x158d5fa3Ef8e4dDA8a5367deCF76b94E7efFCe95;
+
+  /// @dev The address of Matic on Polygon
+  address public constant NATIVE_MATIC = 0x0000000000000000000000000000000000001010;
```
sendra: Not used

scripts/DeployBridges.s.sol
efecarranza: @sendra Yeah, last time they asked that I deployed, do a live test with the deployed contract and transfer ownership. The polygon erc20 bridge has this TX history as well.

src/bridges/polygon/AavePolEthPlasmaBridge.sol
efecarranza: How do you do that? You can't set variables on interfaces.

brotherlymite: @efecarranza The getters for public variables are automatically generated. So on the interface you can add something like: `function ERC20_PREDICATE_BURN() external returns (address);` with natspec, and over here you use inheritdoc for comments.

src/bridges/polygon/AavePolEthPlasmaBridge.sol
```diff
@@ -0,0 +1,103 @@
...
+  /// @param _owner The owner of the contract upon deployment
+  constructor(address _owner) {
+    _transferOwnership(_owner);
+  }
+
+  /// @inheritdoc IAavePolEthPlasmaBridge
+  function bridge(address token, uint256 amount) external onlyOwner {
+    if (block.chainid != ChainIds.POLYGON) revert InvalidChain();
+
+    IERC20Polygon(token).withdraw{value: amount}(amount);
```
brotherlymite: @efecarranza any reason why we are passing `{value: amount}` here. For ex. when we try to bridge 5 AAVE tokens from polygon to mainnet, we should not need to pay 5 MATIC here no?

tests/bridges/polygon/AavePolEthPlasmaBridge.t.sol
```diff
@@ -0,0 +1,208 @@
...
+
+    vm.expectRevert('Withdrawer and burn exit tx do not match');
+    bridgeMainnet.confirmExit(burnProof);
+  }
+}
+
+/// This is a real proof that was manually created by efecarranza.eth
+/// The TX can be found here: 
+contract ForkedBridgeTests is Test {
```
brotherlymite, line 208: missing TODO

src/bridges/polygon/AavePolEthPlasmaBridge.sol
efecarranza: @brotherlymite Matic is the native token of Polygon, it works differently than standard ERC20 tokens. it has to be sent in the value.

efecarranza:
@brotherlymite @sendra updated all !

thanks

src/bridges/polygon/AavePolEthPlasmaBridge.sol
brotherlymite: @efecarranza Maybe when we are bridging MATIC token it could be, but in case where we are bridging any other token - for example 10000 UNI tokens, we don't need to pay 10000 native token (Matic) right?

sendra: so the thing here is, if plasma should only be for matic, why pass token address no? or am i missing something here?

efecarranza: @sendra I was leaving it open in case they open up the bridge in the future, but we can cross that bridge when we get to it and just allow MATIC for now.

tests/bridges/polygon/AavePolEthPlasmaBridge.t.sol
efecarranza: added

efecarranza:
@sendra @brotherlymite updated and only allows MATIC bridges now.

src/bridges/polygon/AavePolEthPlasmaBridge.sol
```diff
@@ -0,0 +1,109 @@
...
+  /// @inheritdoc IAavePolEthPlasmaBridge
+  function exit() external {
+    if (block.chainid != ChainIds.MAINNET) revert InvalidChain();
+
+    IWithdrawManager(WITHDRAW_MANAGER).processExits(MATIC_MAINNET);
+    emit Exit(MATIC_MAINNET);
+  }
+
+  /// @inheritdoc IAavePolEthPlasmaBridge
+  function withdrawToCollector(address token) external {
```
sendra: This should also be using the MATIC_MAINNET constant right? as it can only withdraw matic to collector? so no need for the arg token?

sendra (approved):

sendra (approved):

brotherlymite (approved):

PR #305, opened 2024-06-19
feat: Add base ccc update payload.

sendra:


src/adi/test/ADITestBase.sol
```diff
@@ -382,31 +383,42 @@ contract ADITestBase is Test {
...
   }
 
   function createConfigurationSnapshot(
     string memory reportName,
     address crossChainController,
     bool receiverConfigs,
     bool receiverAdapterConfigs,
-    bool forwarderAdapterConfigs
+    bool forwarderAdapterConfigs,
+    bool cccImplUpdate
```
kyzia551: please replace it with struct

kyzia551 (changes requested):

kyzia551 (approved):

PR #307, opened 2024-07-01
Aave wstETH native withdrawer

defijesus:
What: A smart contract that can start and finalize native stETH withdrawals.

Why: A small contract like this would greatly improve the dev UX when handling native stETH withdrawals.

Deployed at: [0x9f489554d5909F372d80C88229E83dcB6629827a](https://etherscan.io/address/0x9f489554d5909F372d80C88229E83dcB6629827a)

Live tests on permissionless version:
- [startWithdraw](https://etherscan.io/tx/0x2b9f8bdacf4f1416872d840c205a202f90bfffa6d47d64b7b221f02b6976b0bc)
- [finalizeWithdraw](https://etherscan.io/tx/0x46272046fb4182eb28b12a24d861d30ea474e90c774bba702964a17544d51f07)


src/asset-manager/AaveStethWithdrawer.sol
```diff
@@ -0,0 +1,77 @@
...
+
+  /// stores a mapping of index to arrays of requestIds
+  mapping(uint256 => uint256[]) public requestIds;
+
+  IWithdrawalQueueERC721 public constant WSETH_WITHDRAWAL_QUEUE =
+    IWithdrawalQueueERC721(0x889edC2eDab5f40e902b864aD4d7AdE8E412F9B1);
+
+  constructor(address owner) {
+    _transferOwnership(owner);
+    IERC20(AaveV3EthereumAssets.wstETH_UNDERLYING).approve(
```
efecarranza: I would do the approvals on startWithdraw

src/asset-manager/AaveStethWithdrawer.sol
```diff
@@ -0,0 +1,77 @@
...
+      address(AaveV3Ethereum.COLLECTOR),
+      ethBalance
+    );
+
+    emit FinalizedWithdrawal(ethBalance, index);
+  }
+  
+  /// @inheritdoc IAaveStethWithdrawer
+  function getRequestIds(uint256 index) public view returns (uint256[] memory reqIds) {
+    reqIds = requestIds[index];
```
efecarranza: I think `return requestIds[index]` is a bit more readable for such a small function

src/asset-manager/interfaces/IAaveStethWithdrawer.sol
```diff
@@ -0,0 +1,80 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+interface IAaveStethWithdrawer {
+  /// @notice emited when a new Withdrawal is requested
```
efecarranza: emitted, typo

src/asset-manager/interfaces/IAaveStethWithdrawer.sol
```diff
@@ -0,0 +1,80 @@
...
+  /// @param index the storage index of the respective requestIds used to finalize the withdrawal
+  event StartedWithdrawal(uint256[] amounts, uint256 index);
+
+  /// @notice emited when a new Withdrawal is requested
+  /// @param amount the amount of WETH withdrawn to collector
+  /// @param index the storage index of the respective requestIds used to finalize the withdrawal
+  event FinalizedWithdrawal(uint256 amount, uint256 index);
+
+  /// @notice Starts a new withdrawal
+  /// @param amounts a list of amounts to be withdrawn. each amount must be > 100 wei and < 1000 ETH
```
efecarranza: If these are the min/max, I'd add these checks and revert.

tests/asset-manager/AaveStethWithdrawerTest.t.sol
```diff
@@ -0,0 +1,168 @@
...
+    vm.startPrank(EXECUTOR);
+    AaveV3Ethereum.COLLECTOR.transfer(
+      address(WSTETH), 
+      address(withdrawer), 
+      100
+    );
+    uint256[] memory amounts = new uint256[](1);
+    amounts[0] = 100;
+    vm.expectEmit(address(withdrawer));
+    emit StartedWithdrawal(amounts, 0);
```
efecarranza: Shouldn't this be index 1? You're doing nextId++ before assigning

src/asset-manager/interfaces/IAaveStethWithdrawer.sol
defijesus: the requestWithdrawalsWstETH() call to the WSETH_WITHDRAWAL_QUEUE already reverts if amounts are outside bounds

src/asset-manager/AaveStethWithdrawer.sol
defijesus: Doing a single approval for max, greatly cleans up the code of startWithdraw. To do approvals only for the amounts needed we would need to go over the amounts array and sum it. Considering this I still think this design is simpler.

src/asset-manager/AaveStethWithdrawer.sol
```diff
@@ -0,0 +1,77 @@
...
+/// TODO include after Rescuable721 gets merged in
+// import {Rescuable721} from 'solidity-utils/contracts/utils/Rescuable721.sol';
+import {AaveV3Ethereum, AaveV3EthereumAssets} from 'aave-address-book/AaveV3Ethereum.sol';
+import {IAaveStethWithdrawer, IWithdrawalQueueERC721, IWETH} from './interfaces/IAaveStethWithdrawer.sol';
+
+contract AaveStethWithdrawer is Ownable, Rescuable, IAaveStethWithdrawer {
+  using SafeERC20 for IERC20;
+
+  /// auto incrementing index to store requestIds of withdrawals
+  uint256 public nextIndex;
```
kyzia551: I think that`nextRequestIndex` will be a better name, or something about withdraw

But actually it's not next, but current, because you do ++ straight away in the `startWithdraw` func

src/asset-manager/AaveStethWithdrawer.sol
```diff
@@ -0,0 +1,77 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import {IERC20} from 'solidity-utils/contracts/oz-common/interfaces/IERC20.sol';
+import {SafeERC20} from 'solidity-utils/contracts/oz-common/SafeERC20.sol';
+import {Ownable} from 'solidity-utils/contracts/oz-common/Ownable.sol';
+import {Rescuable} from 'solidity-utils/contracts/utils/Rescuable.sol';
+/// TODO include after Rescuable721 gets merged in
+// import {Rescuable721} from 'solidity-utils/contracts/utils/Rescuable721.sol';
```
kyzia551: please update

src/asset-manager/AaveStethWithdrawer.sol
```diff
@@ -0,0 +1,77 @@
...
+import {AaveV3Ethereum, AaveV3EthereumAssets} from 'aave-address-book/AaveV3Ethereum.sol';
+import {IAaveStethWithdrawer, IWithdrawalQueueERC721, IWETH} from './interfaces/IAaveStethWithdrawer.sol';
+
```
kyzia551: Please add credits about author, etc

src/asset-manager/AaveStethWithdrawer.sol
```diff
@@ -0,0 +1,77 @@
...
+    _transferOwnership(owner);
+    IERC20(AaveV3EthereumAssets.wstETH_UNDERLYING).approve(
+      address(WSETH_WITHDRAWAL_QUEUE),
+      type(uint256).max
+    );
+  }
+
+  /// @inheritdoc IAaveStethWithdrawer
+  function startWithdraw(uint256[] calldata amounts) external {
+    uint256 index = nextIndex++;
```
kyzia551: why don't you start from 0?

src/asset-manager/interfaces/IAaveStethWithdrawer.sol
```diff
@@ -0,0 +1,80 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+interface IAaveStethWithdrawer {
+  /// @notice emitted when a new Withdrawal is requested
+  /// @param amounts the amounts requested to be withdrawn
+  /// @param index the storage index of the respective requestIds used to finalize the withdrawal
+  event StartedWithdrawal(uint256[] amounts, uint256 index);
```
kyzia551: make `index` indexed

src/asset-manager/interfaces/IAaveStethWithdrawer.sol
```diff
@@ -0,0 +1,80 @@
...
+
+  /// @notice emitted when a new Withdrawal is requested
+  /// @param amount the amount of WETH withdrawn to collector
+  /// @param index the storage index of the respective requestIds used to finalize the withdrawal
+  event FinalizedWithdrawal(uint256 amount, uint256 index);
```
kyzia551: make `index` indexed

tests/asset-manager/AaveStethWithdrawerTest.t.sol
```diff
@@ -0,0 +1,168 @@
...
+//     );
+//     vm.stopPrank();
+
+//     assertEq(
+//       WSTETH.balanceOf(COLLECTOR),
+//       initialCollectorBalance + 100
+//     );
+//     assertEq(WSTETH.balanceOf(address(withdrawer)), 0);
+//   }
+// }
```
kyzia551: update

kyzia551 (changes requested):

src/asset-manager/AaveStethWithdrawer.sol
```diff
@@ -0,0 +1,96 @@
...
+  /// auto incrementing index to store requestIds of withdrawals
+  uint256 public nextIndex;
+
+  /// stores a mapping of index to arrays of requestIds
+  mapping(uint256 => uint256[]) public requestIds;
+
+  IWithdrawalQueueERC721 public constant WSETH_WITHDRAWAL_QUEUE =
+    IWithdrawalQueueERC721(0x889edC2eDab5f40e902b864aD4d7AdE8E412F9B1);
+
+  constructor(address owner) {
```
eboadom: Any rationale to pass here the executor as parameter, instead of using directly from address book? Considering that this contract is already very coupled to Ethereum with the usage of `AaveV3Ethereum.COLLECTOR` later on

src/asset-manager/AaveStethWithdrawer.sol
```diff
@@ -0,0 +1,96 @@
...
+ * @notice Helper contract to natively withdraw wstETH to the collector
+ */
+contract AaveStethWithdrawer is Ownable, Rescuable721, IAaveStethWithdrawer {
+  using SafeERC20 for IERC20;
+
+  /// auto incrementing index to store requestIds of withdrawals
+  uint256 public nextIndex;
+
+  /// stores a mapping of index to arrays of requestIds
+  mapping(uint256 => uint256[]) public requestIds;
```
eboadom: Given that this is `public` already, probably doesn't make sense to expose the other `getRequestIds()`

src/asset-manager/AaveStethWithdrawer.sol
```diff
@@ -0,0 +1,96 @@
...
+    uint256 index = nextIndex++;
+    uint256[] memory rIds = WSETH_WITHDRAWAL_QUEUE.requestWithdrawalsWstETH(amounts, address(this));
+
+    requestIds[index] = rIds;
+    emit StartedWithdrawal(amounts, index);
+  }
+
+  /// @inheritdoc IAaveStethWithdrawer
+  function finalizeWithdraw(uint256 index) external {
+    uint256[] memory reqIds = getRequestIds(index);
```
eboadom: Minor, but why using the getter instead of the `requestIds` var?

src/asset-manager/AaveStethWithdrawer.sol
```diff
@@ -0,0 +1,96 @@
...
+    uint256[] memory hintIds = WSETH_WITHDRAWAL_QUEUE.findCheckpointHints(
+      reqIds,
+      1,
```
eboadom: My understanding is that `getLastRequestId()` is always incremental, so even if this is only used to limit the binary search, wouldn't make more sense to instead of `1` initialize in a constant/immutable on this contract the value of `getLastRequestId()`?

Because if not it would be the same as using `function claimWithdrawal(uint256 _requestId) external {` later on instead of `claimWithdrawalsTo()`

eboadom:
Some interesting "vector" (don't see any danger, but better to think about it): somebody could donate ETH to the contract while having an active (not withdrawn) stETH order, and `finalizeWithdraw()` would add the amount to withdraw from stETH + the balance of the contract from the donation.

I recommend to add a test dealing with this type of scenario.

eboadom:
Something that I don't fully understand high-level. Wouldn't be reasonable to make use of `onERC721Received()` on `AaveStethWithdrawer`?

tests/asset-manager/AaveStethWithdrawerTest.t.sol
```diff
@@ -0,0 +1,163 @@
...
+  AaveStethWithdrawer public withdrawerReadyToWithdraw;
+
+  function setUp() public {
+    vm.createSelectFork(vm.rpcUrl('mainnet'), 20197457);
+    
+    withdrawer = new AaveStethWithdrawer(EXECUTOR);
+    UNSTETH = IERC20(address(withdrawer.WSETH_WITHDRAWAL_QUEUE()));
+    bytes memory code = address(withdrawer).code;
+
+    /// At current block 0x3169db has an Lido withdrawal NFT
```
eboadom: Is it really needed this complexity instead of reproducing an account withdrawal? This way you are coupling the tests completely to the fork block

src/asset-manager/AaveStethWithdrawer.sol
defijesus: main rational was to not need to update this code when updating the governance contracts to the next version.

from v2 to v3 we saw some changes on the executor side of things, I assume with v4 we'll see some changes too

src/asset-manager/AaveStethWithdrawer.sol
defijesus: I followed the usage recommendation @ the source code of findCheckpointHints here:

https://etherscan.io/address/0xe42c659dc09109566720ea8b2de186c2be7d94d9#code#F22#L293

Will update contract to store an immutable checkpoint on contract creation to prune all older request ids from the binary search

defijesus:
> Something that I don't fully understand high-level. Wouldn't be reasonable to make use of `onERC721Received()` on `AaveStethWithdrawer`?

would be needed if stETH system would call such callback. from my understanding onERC721Received() only triggers if safeTransferFrom is used (which is not the case here)

tests/asset-manager/AaveStethWithdrawerTest.t.sol
defijesus: from my cursory understanding of Lido systems, we would need to mock a lot more calls, and a lot more of Lidos internal systems, to make our test withdrawal available to be finalized.

imo, using the current solution leads to a cleaner tests that focus on testing the actual Withdrawer functionality vs Lido internal workings.

src/asset-manager/AaveStethWithdrawer.sol
eboadom: My point is that even if you want to use a variable owner (to transfer ownership later), I don't see much reason to pass it from outside, because this contract is designed to be one per network, same as Collector

tests/asset-manager/AaveStethWithdrawerTest.t.sol
eboadom: By being invasive (etching, swapping storage) on external systems, it is even difficult to evaluate how close is the test to production. Meaning that maybe will work here and not in production

Take into account that the reviewer (us) don't know in depth that component in Lido, so we simply can't say if this test is legit or not when using cheatcodes. Looks so, but still

eboadom (approved):

PR #326, opened 2024-07-30
CollectorUtils base implementation

kyzia551:


src/CollectorUtils.sol
```diff
@@ -0,0 +1,203 @@
...
+    if (swapperBalance < input.amount) {
+      input.amount = swapperBalance;
+    }
+
+    AaveSwapper(swapper).swap(
+      input.milkman,
+      input.priceChecker,
+      input.fromUnderlying,
+      input.toUnderlying,
+      input.fromUnderlyingPriceFeed,
```
sakulstra, line 171: with recent proposal bugs perhaps good to fetch oracle decimals and ensure they are the same? (probably this should have better been on swapper itself, but as this is to late now, might make sense)

src/CollectorUtils.sol
```diff
@@ -0,0 +1,203 @@
...
+  function __withdraw(
+    ICollector collector,
+    IOInput memory input,
+    address aTokenAddress
+  ) internal returns (uint256) {
+    if (input.amount == 0) {
+      revert InvalidZeroAmount();
+    }
+
+    if (input.amount == type(uint256).max) {
```
sakulstra: not sure if this should be supported as the collector should never be emptied.

src/CollectorUtils.sol
```diff
@@ -0,0 +1,203 @@
...
+      input.amount = IERC20(aTokenAddress).balanceOf(address(collector));
+    }
+    collector.transfer(aTokenAddress, address(this), input.amount);
+
+    // in case of imprecision during the aTokenTransfer withdraw a bit less
+    uint256 balanceAfterTransfer = IERC20(aTokenAddress).balanceOf(address(this));
+    input.amount = balanceAfterTransfer >= input.amount ? input.amount : balanceAfterTransfer;
+
+    // @dev withdrawal interfaces of v2 and v3 is the same, so we use any
+    IPool(input.pool).withdraw(input.underlying, input.amount, address(collector));
```
sakulstra: not getting why this is needed - why not just withdraw `type(uint256).max`?

src/CollectorUtils.sol
kyzia551: Added [here](https://github.com/bgd-labs/aave-helpers/pull/326/files#diff-8405e99f7d86dbbd00abe684f6246d73267e7c0c6db7e9c2efb6e4de17119f78R148-R149)

src/CollectorUtils.sol
kyzia551: in case there will be some funds sent by mistake

src/CollectorUtils.sol
```diff
@@ -0,0 +1,212 @@
...
+    if (input.amount == 0) {
+      revert InvalidZeroAmount();
+    }
+
+    if (input.amount == type(uint256).max) {
+      input.amount = IERC20(aTokenAddress).balanceOf(address(collector));
+    }
+    collector.transfer(aTokenAddress, address(this), input.amount);
+    // @dev withdrawal interfaces of v2 and v3 is the same, so we use any
+    IPool(input.pool).withdraw(input.underlying, type(uint256).max, address(collector));
```
kyzia551: Rephrased
cc: @sakulstra

sakulstra (approved):

PR #338, opened 2024-08-20
fix: remove duplicated code

sakulstra:


zksync/src/SnapshotHelpersV3.sol
```diff
@@ -170,122 +170,6 @@ contract SnapshotHelpersV3 is CommonTestBase, DiffUtils {
     vm.writeJson(output, path);
   }
 
-  function writeReserveConfigs(
-    string memory path,
-    ReserveConfig[] memory configs,
-    IPool pool
-  ) public {
-    _switchOnZkVm();
```
sakulstra, line 178: why is the switching on needed? seems to work fine without.

PR #358, opened 2024-12-05
feat: Financesteward

Luigy-Lemon:


src/financestewards/PoolV3FinSteward.sol
```diff
@@ -0,0 +1,174 @@
...
+import {AaveV2Ethereum, AaveV2EthereumAssets} from 'aave-address-book/AaveV2Ethereum.sol';
+import {IPool, DataTypes as DataTypesV3} from 'aave-address-book/AaveV3.sol';
+import {ILendingPool, DataTypes as DataTypesV2} from 'aave-address-book/AaveV2.sol';
+import {IPoolV3FinSteward} from './interfaces/IPoolV3FinSteward.sol';
+
+/**
+ * @title PoolV3FinSteward
+ * @author luigy-lemon  (Karpatkey)
+ * @author efecarranza  (Tokenlogic)
+ * @notice Helper contract that enables a Guardian to execute permissioned actions on the Aave Collector
```
sakulstra, line 18: Only had a quick look now, but to me this looks still very convoluted & the docs is extremely sparse.
Having a top natspec explaining a bit what is the intention of that contract would go great lengths imo.

By my understanding the intention is:
- to withdraw assets from v2
  - either to swap them to another asset
  - or to migrate them to v3
- to deposit raw underlying to v3 (i assume to account for swap results and things like psp rewards)

So functionality wise this is super narrow:
- funds never leave the collector
- all the actions are in dao interest

---

A lot of the complexity comes from the fact that the DAO wants to keep certain amounts of tokens on the aToken - to account for known deficit and to prevent rounding related issues. Have you thought about:
1. deploying some DAO controlled erc20 token vault
2. move the "immobile" funds there
This way, you will never need to keep track of some amounts any more. I think it's an idea deinitely worth to explore.

src/financestewards/PoolV3FinSteward.sol
```diff
@@ -0,0 +1,174 @@
...
+  using DataTypesV2 for DataTypesV2.ReserveData;
+  using DataTypesV3 for DataTypesV3.ReserveDataLegacy;
+  using CU for ICollector;
+  using CU for CU.IOInput;
+
+  /// @inheritdoc IPoolV3FinSteward
+  ICollector public immutable COLLECTOR = AaveV3Ethereum.COLLECTOR;
+
+  /// @inheritdoc IPoolV3FinSteward
+  ILendingPool public v2Pool;
```
sakulstra: why does this exist?

src/financestewards/PoolV3FinSteward.sol
```diff
@@ -0,0 +1,174 @@
...
+
+  /// @inheritdoc IPoolV3FinSteward
+  mapping(address pool => bool isApproved) public v3Pools;
```
sakulstra:
```suggestion
  mapping(address pool => bool isApproved) public pools;
```

why not just approve pools independent of version?

src/financestewards/PoolV3FinSteward.sol
```diff
@@ -0,0 +1,174 @@
...
+ * @notice Helper contract that enables a Guardian to execute permissioned actions on the Aave Collector
+ */
+contract PoolV3FinSteward is OwnableWithGuardian, IPoolV3FinSteward {
+  using DataTypesV2 for DataTypesV2.ReserveData;
+  using DataTypesV3 for DataTypesV3.ReserveDataLegacy;
+  using CU for ICollector;
+  using CU for CU.IOInput;
+
+  /// @inheritdoc IPoolV3FinSteward
+  ICollector public immutable COLLECTOR = AaveV3Ethereum.COLLECTOR;
```
sakulstra:
```suggestion
  ICollector public immutable COLLECTOR;
```
assign in constructor.

src/financestewards/PoolV3FinSteward.sol
```diff
@@ -0,0 +1,174 @@
...
+
+  constructor(address _owner, address _guardian, address _v2Pool, address[] memory _v3Pools) {
+    _transferOwnership(_owner);
+    _updateGuardian(_guardian);
+    _setV2Pool(_v2Pool);
+
+    for (uint256 i; i < _v3Pools.length; ) {
+      _setV3Pool(_v3Pools[i]); // Main
+      unchecked { ++i; }
+}
```
sakulstra: why is this not linted.

src/financestewards/PoolV3FinSteward.sol
```diff
@@ -0,0 +1,174 @@
...
+  mapping(address token => uint256 minimumBalanceLeft) public minTokenBalance;
+
+  constructor(address _owner, address _guardian, address _v2Pool, address[] memory _v3Pools) {
+    _transferOwnership(_owner);
+    _updateGuardian(_guardian);
+    _setV2Pool(_v2Pool);
+
+    for (uint256 i; i < _v3Pools.length; ) {
+      _setV3Pool(_v3Pools[i]); // Main
+      unchecked { ++i; }
```
sakulstra: let's not do weird gas optimizations for the sake of saving 10 gas if readability suffers.

src/financestewards/PoolV3FinSteward.sol
```diff
@@ -0,0 +1,174 @@
...
+  /// @inheritdoc IPoolV3FinSteward
+  ILendingPool public v2Pool;
+
+  /// @inheritdoc IPoolV3FinSteward
+  mapping(address pool => bool isApproved) public v3Pools;
+
+  /// @inheritdoc IPoolV3FinSteward
+  mapping(address token => uint256 minimumBalanceLeft) public minTokenBalance;
+
+  constructor(address _owner, address _guardian, address _v2Pool, address[] memory _v3Pools) {
```
sakulstra:
```suggestion
  constructor(address _owner, address _guardian, address[] memory _pools) {
```

src/financestewards/PoolV3FinSteward.sol
```diff
@@ -0,0 +1,174 @@
...
+    _transferOwnership(_owner);
+    _updateGuardian(_guardian);
+    _setV2Pool(_v2Pool);
+
+    for (uint256 i; i < _v3Pools.length; ) {
```
sakulstra:
```suggestion
    for (uint256 i; i < _pools.length; ) {
```

src/financestewards/PoolV3FinSteward.sol
```diff
@@ -0,0 +1,174 @@
...
+      _setV3Pool(_v3Pools[i]); // Main
```
sakulstra:
```suggestion
      _approvePool(pools[i]); // Main
```

src/financestewards/PoolV3FinSteward.sol
```diff
@@ -0,0 +1,174 @@
...
+    CU.IOInput memory depositData = CU.IOInput(pool, reserve, amount);
+    CU.depositToV3(COLLECTOR, depositData);
+  }
+
+  /// @inheritdoc IPoolV3FinSteward
+  function withdrawV3(address pool, address reserve, uint256 amount) external onlyOwnerOrGuardian {
+    if (amount == 0) revert InvalidZeroAmount();
+    _validateV3Pool(pool);
+
+    DataTypesV3.ReserveDataLegacy memory reserveData = IPool(pool).getReserveData(reserve);
```
sakulstra: To hook on a comment from above - doing the validation will cost you >10k gas per tx. The loop optimization, will save you essentially nothing. But solving this problem will save actual money on each tx (and each proposal, which faces the same problem).

src/financestewards/PoolV3FinSteward.sol
```diff
@@ -0,0 +1,174 @@
...
+    if (amount == 0) revert InvalidZeroAmount();
+    _validateV3Pool(pool);
+    CU.IOInput memory depositData = CU.IOInput(pool, reserve, amount);
+    CU.depositToV3(COLLECTOR, depositData);
+  }
+
+  /// @inheritdoc IPoolV3FinSteward
+  function withdrawV3(address pool, address reserve, uint256 amount) external onlyOwnerOrGuardian {
+    if (amount == 0) revert InvalidZeroAmount();
+    _validateV3Pool(pool);
```
sakulstra: not sure if a validation would even be needed in that case 🤔 
Withdrawal to the collector i don't see any risk with that (even if it was a foreign pool).

src/financestewards/PoolV3FinSteward.sol
```diff
@@ -0,0 +1,174 @@
...
+
+    _validateV3Pool(pool);
+
+    DataTypesV2.ReserveData memory reserveData = v2Pool.getReserveData(reserve);
+    address atoken = reserveData.aTokenAddress;
+
+    uint256 balanceBefore = IERC20(reserve).balanceOf(address(COLLECTOR));
+    uint256 validAmount = _validateAmount(atoken, amount);
+    CU.IOInput memory withdrawData = CU.IOInput(address(v2Pool), reserve, validAmount);
+    CU.withdrawFromV2(COLLECTOR, withdrawData, address(COLLECTOR));
```
sakulstra: no need for these balance operations - withdraw will return the amount withdrawn i think.

src/financestewards/MainnetSwapSteward.sol
```diff
@@ -0,0 +1,223 @@
...
+
+  /// @inheritdoc ISwapSteward
+  function setSwappableToken(address token, address priceFeedUSD) external onlyOwner {
+    if (priceFeedUSD == address(0)) revert MissingPriceFeed();
+
+    swapApprovedToken[token] = true;
+    priceOracle[token] = priceFeedUSD;
+
+    // Validate oracle has necessary functions
+    AggregatorInterface(priceFeedUSD).decimals();
```
sakulstra: does it make sense to validate also that it's 8?

src/financestewards/MainnetSwapSteward.sol
```diff
@@ -0,0 +1,223 @@
...
+      reserve,
+      buyToken,
+      priceOracle[reserve],
+      priceOracle[buyToken],
+      withdrawAmount,
+      slippage
+    );
+
+    CU.swap(COLLECTOR, address(SWAPPER), swapData);
+  }
```
sakulstra: why is this code actually needed 🤔 , i think is better to just isolate the stewards.
If you want to swap just batch 2 txns:
1. on pool steward withdraw
2. on swap steward swap

src/financestewards/MainnetSwapSteward.sol
```diff
@@ -0,0 +1,223 @@
...
+
+  /// @inheritdoc ISwapSteward
+  function tokenSwap(
+    address sellToken,
+    uint256 amount,
+    address buyToken,
+    uint256 slippage
+  ) external onlyOwnerOrGuardian {
+    POOLV3STEWARD.validateAmount(sellToken, amount);
```
sakulstra:
```suggestion
```
no need to validate any amount as it's underlying. Also as commented on the other steward i think the problem should be tackled at the root. It's far cleaner/easer/cheaper then working around it on every occasion.

src/financestewards/MainnetSwapSteward.sol
```diff
@@ -0,0 +1,223 @@
...
+  uint256 public constant MAX_SLIPPAGE = 1000; // 10%
+
+  /// @inheritdoc ISwapSteward
+  AaveSwapper public immutable SWAPPER = AaveSwapper(MiscEthereum.AAVE_SWAPPER);
+
+  /// @inheritdoc ISwapSteward
+  ICollector public immutable COLLECTOR = AaveV3Ethereum.COLLECTOR;
+
+  /// @inheritdoc ISwapSteward
+  IPoolV3FinSteward public POOLV3STEWARD;
```
sakulstra:
```suggestion
```
imo the stewards should be isolated.

src/financestewards/MainnetSwapSteward.sol
```diff
@@ -0,0 +1,223 @@
...
+contract MainnetSwapSteward is OwnableWithGuardian, ISwapSteward {
+  using DataTypesV2 for DataTypesV2.ReserveData;
+  using DataTypesV3 for DataTypesV3.ReserveDataLegacy;
+  using CU for CU.SwapInput;
+
+  /// @inheritdoc ISwapSteward
+  uint256 public constant MAX_SLIPPAGE = 1000; // 10%
+
+  /// @inheritdoc ISwapSteward
+  AaveSwapper public immutable SWAPPER = AaveSwapper(MiscEthereum.AAVE_SWAPPER);
```
sakulstra: as commented on the other, better initialize immutables on constructor, so you potentially can reuse.

src/financestewards/FinanceSteward.sol
```diff
@@ -0,0 +1,137 @@
...
+import {ICollector, CollectorUtils as CU} from '../CollectorUtils.sol';
+import {IFinanceSteward} from './interfaces/IFinanceSteward.sol';
+
+/**
+ * @title FinanceSteward
+ * @author luigy-lemon  (Karpatkey)
+ * @author efecarranza  (Tokenlogic)
+ * @notice Helper contract that enables a Guardian to execute permissioned actions on the Aave Collector
+ */
+contract FinanceSteward is OwnableWithGuardian, IFinanceSteward {
```
sakulstra, line 16: can you elaborate a bit more on this steward.
What exactly is the intention here? Is there a forum post?

src/financestewards/PoolV3FinSteward.sol
Luigy-Lemon: it doesn't work in V2. That's why i implemented it. It was detected during testing.

PR #386, opened 2025-02-25
Added flashloans to e2e tests

TepNik:


src/ProtocolV3TestBase.sol
```diff
@@ -304,6 +308,31 @@ contract ProtocolV3TestBase is RawProtocolV3TestBase, CommonTestBase {
 
       vm.revertToState(snapshotAfterDeposits);
     }
+
+    // test flashloans
+    if (testAssetConfig.isFlashloanable) {
+      MockFlashLoanReceiver flashLoanReceiver = new MockFlashLoanReceiver();
```
sakulstra: probably better to deploy once globally, no?
```suggestion
```

src/ProtocolV3TestBase.sol
```diff
@@ -39,6 +42,7 @@ struct InterestStrategyValues {
  */
 contract ProtocolV3TestBase is RawProtocolV3TestBase, CommonTestBase {
   using ReserveConfiguration for DataTypes.ReserveConfigurationMap;
+  using PercentageMath for uint256;
   using SafeERC20 for IERC20;
```
sakulstra, line 46:
```suggestion
  MockFlashLoanReceiver flashLoanReceiver = new MockFlashLoanReceiver();
```

src/ProtocolV3TestBase.sol
```diff
@@ -518,6 +547,80 @@ contract ProtocolV3TestBase is RawProtocolV3TestBase, CommonTestBase {
...
+      });
+    }
+
+    uint256 underlyingTokenBalanceOfATokenAfter = IERC20(config.underlying).balanceOf(
+      config.aToken
+    );
+    uint256 debtTokenBalanceOfUserAfter = IERC20(config.variableDebtToken).balanceOf(user);
+
+    if (interestRateMode == 0) {
+      if (config.underlying == AaveV3OptimismAssets.sUSD_UNDERLYING) {
```
sakulstra: this i don't understandt, can you elaborate on why sUSD is special in that case?

TepNik: Here is a reason:
https://github.com/bgd-labs/aave-helpers/blob/027f835922a7b0ddab5acf3f8bc1b7a3b9133014/src/CommonTestBase.sol#L122-L130
https://github.com/bgd-labs/aave-helpers/blob/027f835922a7b0ddab5acf3f8bc1b7a3b9133014/src/CommonTestBase.sol#L90-L97

src/ProtocolV3TestBase.sol
TepNik: Done

src/ProtocolV3TestBase.sol
TepNik: Done

sakulstra (approved):

PR #347, opened 2024-09-30
Gho ccip bridge (#7)

codeprom:
### Overview

The `AaveCcipGhoBridge` smart contract facilitates the secure bridging of GHO tokens across different blockchain networks using Chainlink's Cross-Chain Interoperability Protocol (CCIP). By leveraging Chainlink's decentralized infrastructure, it ensures reliable and transparent cross-chain token transfers while using GHO tokens exclusively for fees.

### Key Components

1. **Router and GHO Addresses**:
   - **`ROUTER`**: The Chainlink CCIP router address for cross-chain communication.
   - **`GHO`**: Address of the GHO token, which the contract bridges across chains.

2. **Bridges Mapping**:
   - A mapping of destination chain selectors (chain identifiers) to corresponding bridge addresses on the destination chains.

### Main Functionalities

1. **Token Transfers**:
   - Facilitates secure GHO token transfers between chains.  
   - Collects GHO tokens from the sender, validates the destination chain and bridge configuration, and transfers the tokens via Chainlink CCIP.  
   - The `checkDestination` modifier ensures that the destination chain and bridge are properly configured before any transfer.

2. **Fee Payment**:
   - Fees are paid exclusively in GHO tokens.  
   - Calculates the required fee for a transfer and deducts it directly from the user's GHO balance.

3. **Cross-Chain Message Handling**:
   - Processes incoming cross-chain messages using the `_ccipReceive` function.  
   - Decodes the message, verifies its authenticity, and releases the corresponding GHO tokens to the recipient on the destination chain.

4. **Quote Transfer**:
   - The `quoteTransfer` function allows users to estimate the GHO fee required for a transfer without executing the transfer.

5. **Setting Destination Bridges**:
   - The contract owner can configure or update the bridge addresses for different destination chains via the `setDestinationBridge` function.

### Security Features

- **Role-Based Access Control**:  
  Uses `AccessControl` to manage permissions:  
  - `DEFAULT_ADMIN_ROLE`: Grants administrative rights.  
  - `BRIDGER_ROLE`: Restricts who can initiate bridging operations.  
  This ensures that only authorized users can perform critical functions.  

- **Destination Validation**:  
  The `checkDestination` modifier ensures that the destination chain and bridge address are properly configured before executing transfers.

- **Token Approval Management**:  
  Securely handles GHO token approvals for the router, mitigating over-approval risks.

- **Rescue Functionality**:  
[...]

PR #443, opened 2025-06-03
CCIP Bridge

efecarranza:
# Changelog

Add CCIP Bridge for Aave to transfer GHO from mainnet to pre-defined networks.
Add fork tests.
Add unit tests.


tests/bridges/ccip/AaveGhoCcipBridgeForkTest.t.sol
```diff
@@ -0,0 +1,752 @@
...
+        Client.EVMTokenAmount[] memory invalidTokenTransfers = arbitrumBridge.getInvalidMessage(message.messageId);
+        assertMessage(message, invalidTokenTransfers);
+        vm.stopPrank();
+    }
+
+    function testFuzz_revertsIf_rateLimitExceeded(uint256 amount) public {
+        vm.selectFork(mainnetFork);
+        uint128 limit = mainnetBridge.getRateLimit(ARBITRUM_CHAIN_SELECTOR);
+
+        vm.assume(amount > limit && amount < 1e32); // made top limit to prevent arithmetic overflow
```
CheyenneAtapour: Might be better to use bound here in case of running with very high fuzz runs, to avoid `vm.assume` rejecting too many inputs

tests/bridges/ccip/AaveGhoCcipBridgeUnitTest.t.sol
```diff
@@ -0,0 +1,236 @@
...
+        MockCCIPRouter mockRouter = new MockCCIPRouter();
+        ccipRouter = IRouterClient(address(mockRouter));
+        mockRouter.setFee(mockFee);
+
+        ERC20Mock mockGho = new ERC20Mock();
+        gho = IERC20(address(mockGho));
+
+        bridge = new AaveGhoCcipBridge(address(mockRouter), address(mockGho), collector, admin);
+
+        vm.startPrank(alice);
```
CheyenneAtapour: small nit: No need to start and stop prank, can just use `vm.prank(alice)`

tests/bridges/ccip/AaveGhoCcipBridgeForkTest.t.sol
efecarranza: good point!

efecarranza:
@CheyenneAtapour updated! thanks

src/bridges/ccip/README.md
miguelmtzinf, line 1: empty

src/bridges/ccip/CCIPReceiver.sol
miguelmtzinf, line 1: is this contract provided by chainlink? wondering if makes sense to move this file and `Client` lib into a dedicated chainlink folder, under the ccip one

src/bridges/ccip/AaveGhoCcipBridge.sol
```diff
@@ -0,0 +1,304 @@
...
+import {IAaveGhoCcipBridge} from "./interfaces/IAaveGhoCcipBridge.sol";
+
+/**
+ * @title AaveGhoCcipBridge
+ * @author TokenLogic
+ * @notice It provides bridging capabilities for the GHO token across networks.
+ *
+ * -- Permissions
+ * The contract implements AccessControl for permissioned functions.
+ * The DEFAULT_ADMIN will always be the respective network's Short Executor (Governance).
```
miguelmtzinf: level 1 instead of short

src/bridges/ccip/AaveGhoCcipBridge.sol
```diff
@@ -0,0 +1,304 @@
...
+import {IRouter} from "./interfaces/IRouter.sol";
+import {IRouterClient} from "./interfaces/IRouterClient.sol";
+import {ITokenPool} from "./interfaces/ITokenPool.sol";
+import {IAaveGhoCcipBridge} from "./interfaces/IAaveGhoCcipBridge.sol";
+
+/**
+ * @title AaveGhoCcipBridge
+ * @author TokenLogic
+ * @notice It provides bridging capabilities for the GHO token across networks.
+ *
```
miguelmtzinf: I d move the spec to a README file

src/bridges/ccip/AaveGhoCcipBridge.sol
```diff
@@ -0,0 +1,304 @@
...
+        if (msg.sender != address(this)) revert OnlySelf();
+        _;
+    }
+
+    /**
+     * @param router The address of the Chainlink CCIP router
+     * @param gho The address of the GHO token
+     * @param collector The address of collector on same chain
+     * @param executor The address of the contract executor
+     */
```
miguelmtzinf:
```suggestion
    /**
     * @dev Constructor
     * @param router The address of the Chainlink CCIP router
     * @param gho The address of the GHO token
     * @param collector The address of the Aave Collector
     * @param initialAdmin The address of the initial admin
     */
```

src/bridges/ccip/AaveGhoCcipBridge.sol
```diff
@@ -0,0 +1,304 @@
...
+    address public immutable GHO_TOKEN;
+
+    /// @inheritdoc IAaveGhoCcipBridge
+    address public immutable ROUTER;
+
+    /// @inheritdoc IAaveGhoCcipBridge
+    address public immutable COLLECTOR;
+
+    /// @inheritdoc IAaveGhoCcipBridge
+    address public immutable EXECUTOR;
```
miguelmtzinf: I d remove this and use a role for `whoCanRescue` or just the default admin role.
It's confusing having 2 access control systems in the same contract: access control and dedicated executor access control for rescue.

src/bridges/ccip/AaveGhoCcipBridge.sol
```diff
@@ -0,0 +1,304 @@
...
+     */
+    function _validateDestinationAndLimit(uint64 chainSelector, uint256 amount) internal view {
+        if (destinations[chainSelector] == address(0)) {
+            revert UnsupportedChain();
+        }
+
+        uint128 limit = _getRateLimit(chainSelector);
+        if (amount > limit) {
+            revert RateLimitExceeded(limit);
+        }
```
miguelmtzinf: not checking for bridge limit on Ethereum. Both things (rate limit and bridge limit) will be checked later when executing the transfer on CCIP side, so we can remove both checks if we wanna rely on them

src/bridges/ccip/AaveGhoCcipBridge.sol
```diff
@@ -0,0 +1,304 @@
...
+    {
+        _validateDestinationAndLimit(chainSelector, amount);
+        Client.EVM2AnyMessage memory message = _buildCCIPMessage(chainSelector, amount, gasLimit, feeToken);
+
+        uint256 fee = IRouterClient(ROUTER).getFee(chainSelector, message);
+
+        if (feeToken == address(0)) {
+            if (msg.value < fee) revert InsufficientFee();
+        } else {
+            IERC20(feeToken).transferFrom(msg.sender, address(this), fee);
```
miguelmtzinf: why we do 2 transfers of GHO instead of 1. Separately, pls use `safeTransfer`

tests/bridges/ccip/AaveGhoCcipBridgeUnitTest.t.sol
```diff
@@ -0,0 +1,235 @@
...
+        ERC20Mock mockGho = new ERC20Mock();
+        gho = IERC20(address(mockGho));
+
+        bridge = new AaveGhoCcipBridge(address(mockRouter), address(mockGho), collector, admin);
+
+        vm.prank(alice);
+        gho.approve(address(bridge), type(uint256).max);
+    }
+
+    function _buildCCIPMessage(uint256 _amount, uint256 _gasLimit, address feeToken)
```
miguelmtzinf, line 43: do we wanna put any restriction on `feeToken`?
what happens if the token is not supported in CCIP? cannot recall when it d fail

src/bridges/ccip/AaveGhoCcipBridge.sol
```diff
@@ -0,0 +1,304 @@
...
+
+    /**
+     * @dev Builds ccip message for token transfer
+     * @param chainSelector The chain selector of the destination chain
+     * @param amount The amount of GHO to transfer
+     * @param gasLimit The gas limit on the destination chain
+     * @param feeToken The address of the fee token (use address(0) for fee in native token)
+     * @return message EVM2AnyMessage to transfer GHO cross-chain
+     */
+    function _buildCCIPMessage(uint64 chainSelector, uint256 amount, uint256 gasLimit, address feeToken)
```
miguelmtzinf: do we wanna put any restriction on `feeToken`?
also, what happens if the token is not supported by CCIP. Cannot recall where it d fail

src/bridges/ccip/AaveGhoCcipBridge.sol
```diff
@@ -0,0 +1,304 @@
...
+        uint256 length = message.destTokenAmounts.length;
+        for (uint256 i = 0; i < length; i++) {
+            IERC20(message.destTokenAmounts[i].token).safeTransfer(COLLECTOR, message.destTokenAmounts[i].amount);
+        }
+
+        emit RecoveredInvalidMessage(messageId);
+    }
+
+    /// @inheritdoc IAaveGhoCcipBridge
+    function setDestinationChain(uint64 chainSelector, address bridge) external onlyRole(DEFAULT_ADMIN_ROLE) {
```
miguelmtzinf: what is the permissions hierarchy / model? Everything is under default admin role...
is just that we wanna have multiple admins?

src/bridges/ccip/AaveGhoCcipBridge.sol
```diff
@@ -0,0 +1,304 @@
...
+        destinations[chainSelector] = bridge;
+
+        emit DestinationChainSet(chainSelector, bridge);
+    }
+
+    /// @inheritdoc IAaveGhoCcipBridge
+    function removeDestinationChain(uint64 chainSelector) external onlyRole(DEFAULT_ADMIN_ROLE) {
+        delete destinations[chainSelector];
+
+        emit DestinationChainRemoved(chainSelector);
```
miguelmtzinf: feels that this can be replaced with `emit DestinationChainSet(chainSelector, address(0));`

src/bridges/ccip/AaveGhoCcipBridge.sol
```diff
@@ -0,0 +1,304 @@
...
+        }
+    }
+
+    /// @inheritdoc IAaveGhoCcipBridge
+    function processMessage(Client.Any2EVMMessage calldata message) external onlySelf {
+        if (destinations[message.sourceChainSelector] != abi.decode(message.sender, (address))) {
+            revert UnknownSourceDestination();
+        }
+
+        _ccipReceive(message);
```
miguelmtzinf: i don't think an internal method is needed here, only used once

src/bridges/ccip/AaveGhoCcipBridge.sol
```diff
@@ -0,0 +1,304 @@
...
+        catch (bytes memory err) {
+            failedMessages[message.messageId] = true;
+            failedTokenTransfers[message.messageId] = message;
+
+            emit FailedToFinalizeBridge(message.messageId, err);
+        }
+    }
+
+    /// @inheritdoc IAaveGhoCcipBridge
+    function processMessage(Client.Any2EVMMessage calldata message) external onlySelf {
```
miguelmtzinf: why this cannot be internal?

src/bridges/ccip/AaveGhoCcipBridge.sol
```diff
@@ -0,0 +1,304 @@
...
+
+        return messageId;
+    }
+
+    /// @inheritdoc CCIPReceiver
+    function ccipReceive(Client.Any2EVMMessage calldata message) external override onlyRouter {
+        try this.processMessage(message) {}
+        catch (bytes memory err) {
+            failedMessages[message.messageId] = true;
+            failedTokenTransfers[message.messageId] = message;
```
miguelmtzinf: we got feedback from Chainlink saying that it's not needed to store the entire msg, can you revisit that please?

src/bridges/ccip/AaveGhoCcipBridge.sol
CheyenneAtapour: (and the last param in the constructor should change to match ofc)

src/bridges/ccip/AaveGhoCcipBridge.sol
miguelmtzinf: ah, was assuming we only work with GHO as fee token if not ETH.

src/bridges/ccip/AaveGhoCcipBridge.sol
miguelmtzinf: reference: https://github.com/bgd-labs/aave-helpers/pull/347#discussion_r1877949626

miguelmtzinf:
Reference of old PR: https://github.com/bgd-labs/aave-helpers/pull/347

src/bridges/ccip/AaveGhoCcipBridge.sol
efecarranza: what would you return in `whoCanRescue()` for the address? didn't understand that part.
something like:

```
whoCanRescue() {
if (hasRole(DEFAULT_ADMIN, msg.sender) return msg.sender
``` 

?

src/bridges/ccip/AaveGhoCcipBridge.sol
efecarranza: didn't understand the first part of `not checking for bridge limit on Ethereum.`

src/bridges/ccip/AaveGhoCcipBridge.sol
efecarranza: the CCIP contract is validating this already, have a test for it here: https://github.com/bgd-labs/aave-helpers/pull/443/files#diff-7dc455c50fa609f9da307dc1b4b6fa831fb5e8d6f57f2d8e1a21effc7920df91R323

src/bridges/ccip/AaveGhoCcipBridge.sol
efecarranza: to me, these set functions should only be done via governance, which would be the admin.
for bridging, i'd allow a guardian (bridger role) to call that so it can be assigned to an allowed party like the aave finance committee

src/bridges/ccip/AaveGhoCcipBridge.sol
efecarranza: it's overriding the CCIPReceiver.sol abstract contract

src/bridges/ccip/AaveGhoCcipBridge.sol
efecarranza: try/catch on solidity only works for external methods

src/bridges/ccip/AaveGhoCcipBridge.sol
```diff
@@ -0,0 +1,293 @@
...
+
+    /// @inheritdoc CCIPReceiver
+    function ccipReceive(Client.Any2EVMMessage calldata message) external override onlyRouter {
+        try this.processMessage(message) {}
+        catch (bytes memory err) {
+            failedMessages[message.messageId] = true;
+
+            uint256 amountsLength = message.destTokenAmounts.length;
+            for (uint256 i = 0; i < amountsLength; i++) {
+                failedTokenTransfers[message.messageId].push(message.destTokenAmounts[i]);
```
miguelmtzinf: cannot we store the ccip message id instead? It d be simpler and going through a loop feels bad.

src/bridges/ccip/AaveGhoCcipBridge.sol
miguelmtzinf: it seems you gotta use `RescuableACL` instead and add proper ACL checks to `_checkRescueGuardian`. We can use DEFAULT_ADMIN or an ad-hoc role, no strong opinion

src/bridges/ccip/AaveGhoCcipBridge.sol
miguelmtzinf: a bit of literature about GHO CCIP and bridge limit: https://www.llamarisk.com/research/explainer-series-ccip
I d suggest to remove rate limit check and keep this simple

src/bridges/ccip/interfaces/IAaveGhoCcipBridge.sol
```diff
@@ -0,0 +1,198 @@
...
+     * @param bridge The address of the bridge on the destination chain
+     */
+    event DestinationChainSet(uint64 indexed chainSelector, address indexed bridge);
+
+    /**
+     * @dev Emitted when an invalid message is received by the bridge
+     * @param messageId The ID of message
+     * @param err The error on why the transfer failed
+     */
+    event FailedToFinalizeBridge(bytes32 indexed messageId, bytes err);
```
miguelmtzinf: I suggest to rename all events related to messages:
`BridgeMessageInitiated`, `BridgeMessageFinalized`, `BridgeMessageFailed` and `BrdigeMessageRecovered`

src/bridges/ccip/interfaces/IAaveGhoCcipBridge.sol
```diff
@@ -0,0 +1,198 @@
...
+     * @dev Emits when receive invalid message
+     * @param messageId The id of message
+     */
+    event RecoveredInvalidMessage(bytes32 indexed messageId);
+
+    /**
+     * @notice This role defines which users can call the send() function.
+     * @return The bytes32 role identifier
+     */
+    function BRIDGER_ROLE() external view returns (bytes32);
```
miguelmtzinf: read-only functions should be placed after write functions, per [solidity style guideline](https://docs.soliditylang.org/en/latest/style-guide.html#order-of-layout)

src/bridges/ccip/README.md
```diff
@@ -0,0 +1,22 @@
+# AaveGhoCcipBridge
+
+The AaveGhoCcipBridge is a contract to facilitate moving GHO from Ethereum Mainnet to various L2 networks utilizing the underlying CCIP infrastructure from Chainlink.
```
miguelmtzinf:
```suggestion
The AaveGhoCcipBridge is a contract to facilitate moving GHO from Ethereum Mainnet to other networks utilizing the underlying CCIP infrastructure from Chainlink.
```

src/bridges/ccip/AaveGhoCcipBridge.sol
efecarranza: leaving here for posterity per Chainlink's devs:

"That's not safe. The person calling that function can just submit any token/amount they want. We don't expose full messages: we never store them as its absurdly expensive"

"I think that makes sense. Looks very much like something we wrote that does the same.
Storing the tokens and amounts makes sense, you don't need the entire msg. You could store the receiver and send the funds to them, assuming it's an EOA or a contract that can just straight up receive funds. But the solution above works in all cases, which is nice. It should be rare to invoke anyway."

src/bridges/ccip/chainlink/libraries/Client.sol
```diff
@@ -0,0 +1,55 @@
+// SPDX-License-Identifier: MIT
```
brotherlymite: i would move all ccip contracts and interfaces except the main bridge contract to dependencies directory

src/bridges/ccip/AaveGhoCcipBridge.sol
```diff
@@ -0,0 +1,293 @@
...
+import {IRouterClient} from "./interfaces/IRouterClient.sol";
+import {ITokenPool} from "./interfaces/ITokenPool.sol";
+import {IAaveGhoCcipBridge} from "./interfaces/IAaveGhoCcipBridge.sol";
+
+/**
+ * @title AaveGhoCcipBridge
+ * @author TokenLogic
+ * @notice It provides bridging capabilities for the GHO token across networks.
+ */
+contract AaveGhoCcipBridge is CCIPReceiver, AccessControl, Rescuable, IAaveGhoCcipBridge {
```
brotherlymite: @efecarranza can you run `npm run lint:fix` to fix lint

src/bridges/ccip/AaveGhoCcipBridge.sol
brotherlymite: agree on this, we validate the rateLimit on TokenPool contract already so cleaner to remove from here

src/bridges/ccip/AaveGhoCcipBridge.sol
```diff
@@ -0,0 +1,293 @@
...
+        Client.EVMTokenAmount[] memory tokenAmounts = new Client.EVMTokenAmount[](1);
+        tokenAmounts[0] = Client.EVMTokenAmount({token: GHO_TOKEN, amount: amount});
+
+        Client.EVM2AnyMessage memory message = Client.EVM2AnyMessage({
+            receiver: abi.encode(destinations[chainSelector]),
+            data: "",
+            tokenAmounts: tokenAmounts,
+            extraArgs: gasLimit == 0
+                ? bytes("")
+                : Client._argsToBytes(Client.EVMExtraArgsV2({gasLimit: gasLimit, allowOutOfOrderExecution: false})),
```
brotherlymite: could we have a network where we would need `allowOutOfOrderExecution` to be set to true?

src/bridges/ccip/AaveGhoCcipBridge.sol
```diff
@@ -0,0 +1,293 @@
...
+    {
+        _validateDestinationAndLimit(chainSelector, amount);
+        Client.EVM2AnyMessage memory message = _buildCCIPMessage(chainSelector, amount, gasLimit, feeToken);
+
+        uint256 fee = IRouterClient(ROUTER).getFee(chainSelector, message);
+
+        if (feeToken == address(0)) {
+            if (msg.value < fee) revert InsufficientFee();
+        } else {
+            IERC20(feeToken).transferFrom(msg.sender, address(this), fee);
```
brotherlymite: better to use safeTransferFrom no? as fee token could be anything or change

src/bridges/ccip/chainlink/libraries/Client.sol
efecarranza: @brotherlymite where would you create that directory? at what level i mean

brotherlymite: over [here](https://github.com/bgd-labs/aave-helpers/tree/main/src/dependencies)

efecarranza:
@brotherlymite @CheyenneAtapour finally got a response from chainlink on out of order execution:

```
Unless your use case requires strict sequencing (e.g., maintaining state updates in order), OoO execution is generally ok for most token transfer scenarios. If you know for sure you don't have a need for strict sequencing of execution of cross-chain transfers at the address level, you could avoid having to configure this for each chain and set OoO to true by default. But only if you are certain, if it's not to difficult then set it up so that it's configurable per chain and you can setting OoO to true only when required, in an abundance of caution. There are only a minority of chains that OoO execution is required to be allowed because of how the chain works. For the most possible assurance we'd recommend this be in scope of an audit.
```

so i made it configurable and also pushed the rest of the changes. this should be it and then once the code is settled, i'll do a live test and add it here as proof of concept before closing.

efecarranza:
@brotherlymite @CheyenneAtapour 

one question i have is about the final destintaion which currently is hardcoded as the COLLECTOR.

with the idea that we fund the Remote GSM: https://github.com/aave-dao/gho-origin/pull/1 via a facilitator and this bridge, the flow would be:

via governance:
1. mint GHO with the OwnableFacilitator
2. bridge to new network

arrives at destination and now we need a new payload to send the GHO to the GhoReserve, this is fine since we can stop the bots and then just execute once the GHO is in the collector but wanted to make sense if we also could potentially add a GhoReserve as a possible destination in the bridge so it goes directly there without the need for a payload.

thank you

efecarranza:
After further discussions with the BGD team, since the funding of the GSM is a one-time scenario, or not very likely, there is no need to make this code more complex and it's okay to do the payload to transfer the funds from the GSM once received in the L2 Collector.

src/bridges/ccip/interfaces/IAaveGhoCcipBridge.sol
```diff
@@ -0,0 +1,230 @@
...
+    uint64 indexed destinationChainSelector,
+    address indexed from,
+    uint256 amount
+  );
+
+  /**
+   * @dev Emitted when the token transfer is executed on the destination chain
+   * @param messageId The ID of the cross-chain message
+   * @param to The address of receiver on destination chain
+   * @param amount The amount of token to translated
```
CheyenneAtapour: Typo here?

```suggestion
   * @param amount The amount of token transferred
```

src/bridges/ccip/interfaces/IAaveGhoCcipBridge.sol
```diff
@@ -0,0 +1,230 @@
...
+   * @dev The executor has the DEFAULT_ADMIN_ROLE
+   * @return The address of the Executor contract
+   */
+  function EXECUTOR() external view returns (address);
+
+  /**
+   * @notice Returns the address of the corresponding bridge for a specified chain selector.
+   * It also returns whether out of order execution is allowed.
+   * @param chainSelector The chain selector of destination chain
+   * @return The struct containing AaveGhoCcipBridge information on destination chain
```
CheyenneAtapour: This comment can be misleading, because it doesn't return a struct, but rather the 2 values, address and bool

src/bridges/ccip/interfaces/IAaveGhoCcipBridge.sol
```diff
@@ -0,0 +1,230 @@
...
+  /**
+   * @notice Returns the executor (governance) address on the deployed chain
+   * @dev The executor has the DEFAULT_ADMIN_ROLE
+   * @return The address of the Executor contract
+   */
+  function EXECUTOR() external view returns (address);
+
+  /**
+   * @notice Returns the address of the corresponding bridge for a specified chain selector.
+   * It also returns whether out of order execution is allowed.
```
CheyenneAtapour: Can simplify this notice into a single sentence, e.g. `Returns the address of the corresponding bridge for a specified chain selector, and whether out of order execution is allowed.`

src/bridges/ccip/AaveGhoCcipBridge.sol
```diff
@@ -0,0 +1,330 @@
...
+
+  /// @inheritdoc IAaveGhoCcipBridge
+  mapping(uint64 chainSelector => Destinations bridgeInfo) public destinations;
+
+  /// @dev Map containing failed token transfer amounts for a message
+  mapping(bytes32 messageId => Client.EVMTokenAmount[] destTokenAmounts)
+    private failedTokenTransfers;
+
+  /// @dev Map containing failed messages and their status
+  mapping(bytes32 messageId => bool isFailed) private failedMessages;
```
CheyenneAtapour: Small nit: I'm personally in favor of prefixing private / internal variables with underscore, e.g. `_failedTokenTransfers`

tests/bridges/ccip/AaveGhoCcipBridgeForkTest.t.sol
```diff
@@ -0,0 +1,871 @@
...
+import {AaveV3Arbitrum, AaveV3ArbitrumAssets} from 'aave-address-book/AaveV3Arbitrum.sol';
+import {GhoArbitrum} from 'aave-address-book/GhoArbitrum.sol';
+import {GovernanceV3Arbitrum} from 'aave-address-book/GovernanceV3Arbitrum.sol';
+import {IRescuable} from 'solidity-utils/contracts/utils/interfaces/IRescuable.sol';
+
+import {CCIPLocalSimulatorFork, Register, Internal} from './mocks/CCIPLocalSimulatorFork.sol';
+import {Constants} from './Constants.sol';
+import {Client} from 'src/dependencies/chainlink/libraries/Client.sol';
+import {CCIPReceiver} from 'src/dependencies/chainlink/CCIPReceiver.sol';
+import {IGhoToken} from './IGhoToken.sol';
```
CheyenneAtapour: small nit to use absolute paths on all imports for consistency, but it's nbd because it's a test file

tests/bridges/ccip/AaveGhoCcipBridgeForkTest.t.sol
```diff
@@ -0,0 +1,871 @@
...
+    );
+    assertMessage(message, invalidTokenTransfers);
+    vm.stopPrank();
+  }
+
+  function testFuzz_revertsIf_rateLimitExceeded(uint256 amount) public {
+    vm.selectFork(mainnetFork);
+    uint128 limit = mainnetBridge.getRateLimit(ARBITRUM_CHAIN_SELECTOR);
+
+    // vm.assume(amount > limit && amount < 1e32); // made top limit to prevent arithmetic overflow
```
CheyenneAtapour: You can remove the commented-out assume code and perhaps move the comment to describe the bound line below

tests/bridges/ccip/AaveGhoCcipBridgeForkTest.t.sol
```diff
@@ -0,0 +1,871 @@
...
+    vm.expectRevert(abi.encodeWithSelector(IAaveGhoCcipBridge.RateLimitExceeded.selector, limit));
+    mainnetBridge.send{value: 100}(ARBITRUM_CHAIN_SELECTOR, amount, 0, feeToken);
+    vm.stopPrank();
+  }
+
+  function test_successful_fuzz(uint256 amount) public {
+    vm.selectFork(mainnetFork);
+    uint128 limit = mainnetBridge.getRateLimit(ARBITRUM_CHAIN_SELECTOR);
+
+    vm.assume(amount > 0 && amount <= limit);
```
CheyenneAtapour: maybe just better to bound amount to avoid vm.assume rejecting too many inputs if increasing fuzz runs

tests/bridges/ccip/AaveGhoCcipBridgeForkTest.t.sol
```diff
@@ -0,0 +1,871 @@
...
+    );
+    assertEq(afterBalance, beforeBalance + amount, 'Bridged amount not updated correctly');
+    vm.stopPrank();
+  }
+
+  function test_successful_payWithEth(uint256 amount) public {
+    vm.selectFork(mainnetFork);
+    uint128 limit = mainnetBridge.getRateLimit(ARBITRUM_CHAIN_SELECTOR);
+
+    vm.assume(amount > 0 && amount <= limit);
```
CheyenneAtapour: Can change any and all of these `vm.assume` statements to use `bound` instead

tests/bridges/ccip/AaveGhoCcipBridgeForkTest.t.sol
```diff
@@ -0,0 +1,871 @@
...
+
+contract SendArbitrumToMainnet is AaveGhoCcipBridgeForkTestBase {
+  address public feeToken = AaveV3ArbitrumAssets.GHO_UNDERLYING;
+
+  function test_revertsIf_unsupportedChain() public {
+    vm.selectFork(arbitrumFork);
+    vm.startPrank(facilitator);
+    vm.expectRevert(IAaveGhoCcipBridge.UnsupportedChain.selector);
+    arbitrumBridge.send(BLAST_CHAIN_SELECTOR, AMOUNT_TO_SEND, 0, feeToken);
+  }
```
CheyenneAtapour, line 359: For consistency, perhaps ensure to stop pranks

tests/bridges/ccip/AaveGhoCcipBridgeForkTest.t.sol
```diff
@@ -0,0 +1,871 @@
...
+  function test_revertIf_messageNotFound() public {
+    vm.selectFork(mainnetFork);
+    vm.startPrank(admin);
+
+    vm.expectRevert(IAaveGhoCcipBridge.MessageNotFound.selector);
+    mainnetBridge.recoverFailedMessageTokens(bytes32(0));
+    vm.stopPrank();
+  }
+
+  function test_successful() public {
```
CheyenneAtapour: small nit, because it's clear from the contract name, but could consider naming this something more descriptive, like `test_successfulMessageRecovery()`

tests/bridges/ccip/mocks/CallWithExactGas.sol
```diff
@@ -0,0 +1,162 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+/// @notice This library contains various callWithExactGas functions. All of them are
+/// safe from gas bomb attacks.
+/// @dev There is code duplication in this library. This is done to not leave the assembly
+/// the blocks.
```
CheyenneAtapour: typo? Just should be `This is done to not leave the assembly blocks.` ?

tests/bridges/ccip/mocks/CallWithExactGas.sol
```diff
@@ -0,0 +1,162 @@
...
+      }
+      g := sub(g, gasForCallExactCheck)
+      // if g - g//64 <= gasAmount, revert. We subtract g//64 because of EIP-150
+      if iszero(gt(sub(g, div(g, 64)), gasLimit)) {
+        mstore(0x0, NOT_ENOUGH_GAS_FOR_CALL_SIG)
+        revert(0x0, 0x4)
+      }
+
+      // call and return whether we succeeded. ignore return data
+      // call(gas,addr,value,argsOffset,argsLength,retOffset,retLength)
```
CheyenneAtapour: probably can delete this comment

src/bridges/ccip/AaveGhoCcipBridge.sol
```diff
@@ -0,0 +1,330 @@
...
+      }
+
+      emit BridgeMessageFailed(message.messageId, err);
+    }
+  }
+
+  /// @inheritdoc IAaveGhoCcipBridge
+  function processMessage(Client.Any2EVMMessage calldata message) external onlySelf {
+    if (
+      destinations[message.sourceChainSelector].destination != abi.decode(message.sender, (address))
```
RensR: This assumes only EVM <> EVM, might be more flexible if you compare the bytes without assuming it's an EVM address. Something like `keccak256(destinations[message.sourceChainSelector].destination) == keccak256(message.sender)`. Or hash destinations[message.sourceChainSelector].destination` on setting the config for reduces gas usage but also reduces visibility into what it means.

src/bridges/ccip/AaveGhoCcipBridge.sol
```diff
@@ -0,0 +1,330 @@
...
+        ),
+      feeToken: feeToken
+    });
+
+    return message;
+  }
+
+  /// @inheritdoc CCIPReceiver
+  function _ccipReceive(Client.Any2EVMMessage memory message) internal override {
+    uint256 ghoAmount = message.destTokenAmounts[0].amount;
```
RensR: imo makes more sense to have this logic in `processMessage`, should also be more gas efficient to not copy the entire msg

src/bridges/ccip/AaveGhoCcipBridge.sol
```diff
@@ -0,0 +1,330 @@
...
+    emit BridgeMessageFinalized(message.messageId, COLLECTOR, ghoAmount);
+  }
+
+  /**
+   * @dev Returns the rate limit of a specified chain.
+   * @param chainSelector The chain selector of the destination chain
+   * @return The rate limit of the chain
+   */
+  function _getRateLimit(uint64 chainSelector) internal view returns (uint128) {
+    address onRamp = IRouter(ROUTER).getOnRamp(chainSelector);
```
RensR, line 290: This works if you don't want a dependency on the TokenAdminRegistry, but that's the official source of truth for pools. This Router call is indirectly calling the Router -> OnRamp -> TokenAdminRegistry. Might still be worth it as you don't need to worry about the TokenAdminRegistry, but that address is considered as stable as the Router so you could add it. Purely a gas improvement as this method also works

src/bridges/ccip/AaveGhoCcipBridge.sol
```diff
@@ -0,0 +1,330 @@
...
+    Client.EVM2AnyMessage memory message = Client.EVM2AnyMessage({
+      receiver: abi.encode(destinationInfo.destination),
+      data: '',
+      tokenAmounts: tokenAmounts,
+      extraArgs: gasLimit == 0
+        ? bytes('')
+        : Client._argsToBytes(
+          Client.EVMExtraArgsV2({
+            gasLimit: gasLimit,
+            allowOutOfOrderExecution: destinationInfo.allowOutOfOrderExecution
```
RensR: I don't see why you would ever set OOO to false here. OOO gives you a nice gas cost reduction (both on sending and exec) and has fewer failure cases. It would also reduce config for this contract if you always turn it on.

src/bridges/ccip/AaveGhoCcipBridge.sol
```diff
@@ -0,0 +1,330 @@
...
+      revert InvalidZeroAmount();
+    }
+
+    Destinations memory destinationInfo = destinations[chainSelector];
+
+    Client.EVMTokenAmount[] memory tokenAmounts = new Client.EVMTokenAmount[](1);
+    tokenAmounts[0] = Client.EVMTokenAmount({token: GHO_TOKEN, amount: amount});
+
+    Client.EVM2AnyMessage memory message = Client.EVM2AnyMessage({
+      receiver: abi.encode(destinationInfo.destination),
```
RensR: If this is just raw bytes you enable non-evm support without further changes, as long as the non-evm supports the same ExtraArgs. A chain that does it Aptos, so you get free Aptos support. Some chains require other extraArgs (Solana). You could consider changing the OOO option to a raw bytes field that is optional. It would hold the encoded extraArgs. If the bytes is empty, fall back to the existing extraArgs logic, if not empty, use the bytes.

src/bridges/ccip/AaveGhoCcipBridge.sol
```diff
@@ -0,0 +1,330 @@
...
+      data: '',
+      tokenAmounts: tokenAmounts,
+      extraArgs: gasLimit == 0
+        ? bytes('')
+        : Client._argsToBytes(
+          Client.EVMExtraArgsV2({
```
RensR: FYI these are renamed GenericExtraArgs in the newer libraries, but are 1:1 the same as EVMExtraArgs. This is to signal they also work for (some) non-EVMS

src/bridges/ccip/AaveGhoCcipBridge.sol
```diff
@@ -0,0 +1,330 @@
...
+    Destinations memory destinationInfo = destinations[chainSelector];
+
+    Client.EVMTokenAmount[] memory tokenAmounts = new Client.EVMTokenAmount[](1);
+    tokenAmounts[0] = Client.EVMTokenAmount({token: GHO_TOKEN, amount: amount});
+
+    Client.EVM2AnyMessage memory message = Client.EVM2AnyMessage({
+      receiver: abi.encode(destinationInfo.destination),
+      data: '',
+      tokenAmounts: tokenAmounts,
+      extraArgs: gasLimit == 0
```
RensR: If it's 0 you cannot use the default! Default gas limit is 200k, not 0. Also as stated in another comment I'd always set OOO.

src/bridges/ccip/AaveGhoCcipBridge.sol
efecarranza: Discussing internally if we want to support non-EVM for this first instance.

src/bridges/ccip/AaveGhoCcipBridge.sol
efecarranza: Since this contract inherits from `CCIPReceiver`, i have to implement this function. Would you just have it empty? Or just emit the event?

src/bridges/ccip/AaveGhoCcipBridge.sol
efecarranza: I'll keep as is for now, this is a governance function that will get called once/twice a year per chain.

src/bridges/ccip/AaveGhoCcipBridge.sol
efecarranza: The first time we were told to keep it as false and then true for chains where it was needed (like ZKSync) but if it's best I'll update.

src/bridges/ccip/AaveGhoCcipBridge.sol
efecarranza: By raw-bytes do you mean `abi.encodePacked()` the receiver instead? 
Didn't understand the Solana option

src/bridges/ccip/AaveGhoCcipBridge.sol
efecarranza: Updated accordingly.

src/bridges/ccip/AaveGhoCcipBridge.sol
efecarranza: I renamed in case we want to support Aptos/other non-EVM.

src/bridges/ccip/interfaces/IAaveGhoCcipBridge.sol
efecarranza: Thanks!

tests/bridges/ccip/mocks/CallWithExactGas.sol
```diff
@@ -0,0 +1,162 @@
...
+      // if g - g//64 <= gasAmount, revert. We subtract g//64 because of EIP-150
+      if iszero(gt(sub(g, div(g, 64)), gasLimit)) {
+        mstore(0x0, NOT_ENOUGH_GAS_FOR_CALL_SIG)
+        revert(0x0, 0x4)
+      }
+
+      // We save the gas before the call so we can calculate how much gas the call used
+      let gasBeforeCall := gas()
+      // call and return whether we succeeded. ignore return data
+      // call(gas,addr,value,argsOffset,argsLength,retOffset,retLength)
```
CheyenneAtapour: can remove this comment

efecarranza:
@RensR I've updated the low-hanging fruit accordingly but have two questions:

1. Remove _ccipReceive(), I have to implement the function.
2. How do I future proof this so it works potentially with Solana/other non-EVM if possible?

src/bridges/ccip/interfaces/IAaveGhoCcipBridge.sol
```diff
@@ -0,0 +1,213 @@
...
+
+  /**
+   * @dev Emits when receive invalid message
+   * @param messageId The id of message
+   */
+  event BridgeMessageRecovered(bytes32 indexed messageId);
+
+  /**
+   * @notice Transfers tokens to specified destination chain.
+   * @dev chain selector can be found https://docs.chain.link/ccip/supported-networks/v1_2_0/mainnet
```
RensR: https://docs.chain.link/ccip/directory/mainnet

This is a better link and it's not versioned so it will always be up to date

RensR:
> @RensR I've updated the low-hanging fruit accordingly but have two questions:
> 
> 1. Remove _ccipReceive(), I have to implement the function.

You can define an empty function and just don't call it. The compiler will remove it as it's internal. 

> 2. How do I future proof this so it works potentially with Solana/other non-EVM if possible?

Storing the entire extraArgs as bytes will give you maximum flexibility, as that would even allow you to add future chain families. It does mean you are no longer able to override the gas limit for those special extraArgs. You could still have the same logic as you currently have when no custom extraArgs bytes are set (create the generic args). 

I actually had a question around the gas limit. Is this code expected to just call this exact contract on some destination chain? If so, the gas required should be known in advance, and there should be no need to ever pass in the gas limit from a user. Currently `send` takes a gas limit, but I don't think that's correct. The dapp would be responsible for setting a correct gas limit, as ccipReceive would require some known gas amount. If you fix that, you simplify the code and you are fully non-evm ready. For e.g. Solana extra args you just encode the args you want offchain and set them in the config, including the gas limit.

src/bridges/ccip/interfaces/IAaveGhoCcipBridge.sol
```diff
@@ -0,0 +1,213 @@
...
+   */
+  function recoverFailedMessageTokens(bytes32 messageId) external;
+
+  /**
+   * @notice Sets a destination chain and corresponding bridge address.
+   * @dev Only callable by ADMIN.
+   * @param chainSelector The chain selector of the destination chain
+   * @param bridge The address of the bridge on the destination chain
+   */
+  function setDestinationChain(uint64 chainSelector, address bridge) external;
```
CheyenneAtapour: Seems BGD wanted more specificity here on natspec, see comment below. Perhaps can clarify with an `@dev` comment that the bridge refers to a contract deployed by us

https://github.com/bgd-labs/aave-helpers/pull/347#discussion_r1818723857

tests/bridges/ccip/AaveGhoCcipBridgeForkTest.t.sol
```diff
@@ -0,0 +1,898 @@
...
+      amount
+    );
+    ccipLocalSimulatorFork.switchChainAndRouteMessage(mainnetFork);
+    uint256 afterBalance = IERC20(AaveV3EthereumAssets.GHO_UNDERLYING).balanceOf(
+      address(AaveV3Ethereum.COLLECTOR)
+    );
+    assertEq(afterBalance, beforeBalance + amount, 'Bridged amount not updated correctly');
+    vm.stopPrank();
+  }
+}
```
CheyenneAtapour, line 594: It was previously requested to add a test where we send extra fee, can you implement this, or point me to the existing test? See below comment:

https://github.com/bgd-labs/aave-helpers/pull/347#discussion_r1819229121

src/bridges/ccip/README.md
```diff
@@ -0,0 +1,22 @@
+# AaveGhoCcipBridge
+
+The AaveGhoCcipBridge is a contract to facilitate moving GHO from Ethereum Mainnet to other networks utilizing the underlying CCIP infrastructure from Chainlink.
```
CheyenneAtapour: Can you please specify in this README that the funds go to `COLLECTOR`?

Originally requested here: https://github.com/bgd-labs/aave-helpers/pull/347#discussion_r1820517138

src/bridges/ccip/AaveGhoCcipBridge.sol
```diff
@@ -0,0 +1,324 @@
...
+import {RescuableBase, IRescuableBase} from 'solidity-utils/contracts/utils/RescuableBase.sol';
+
+import {Client} from '../../dependencies/chainlink/libraries/Client.sol';
+import {CCIPReceiver} from '../../dependencies/chainlink/CCIPReceiver.sol';
+import {IAny2EVMMessageReceiver} from '../../dependencies/chainlink/interfaces/IAny2EVMMessageReceiver.sol';
+import {IOnRampClient} from '../../dependencies/chainlink/interfaces/IOnRampClient.sol';
+import {IRouter} from '../../dependencies/chainlink/interfaces/IRouter.sol';
+import {IRouterClient} from '../../dependencies/chainlink/interfaces/IRouterClient.sol';
+import {ITokenPool} from '../../dependencies/chainlink/interfaces/ITokenPool.sol';
+import {IAaveGhoCcipBridge} from './interfaces/IAaveGhoCcipBridge.sol';
```
CheyenneAtapour: imo better to use absolute paths rather than relative for imports

src/bridges/ccip/AaveGhoCcipBridge.sol
```diff
@@ -0,0 +1,324 @@
...
+          Client.GenericExtraArgs({gasLimit: gasLimit, allowOutOfOrderExecution: true})
+        ),
+      feeToken: feeToken
+    });
+
+    return message;
+  }
+
+  /// @inheritdoc CCIPReceiver
+  function _ccipReceive(Client.Any2EVMMessage memory message) internal override {
```
CheyenneAtapour: According to Solidity style guide (https://docs.soliditylang.org/en/latest/style-guide.html#order-of-functions), view functions should be last within a grouping, but this is mixed in with `internal view`, so please place this function right after `function supportsInterface()`

src/bridges/ccip/interfaces/IAaveGhoCcipBridge.sol
```diff
@@ -0,0 +1,213 @@
...
+   * @notice Returns the Chainlink CCIP router address
+   * @return The address of the Chainlink CCIP router
+   */
+  function ROUTER() external view returns (address);
+
+  /**
+   * @notice Returns the GHO token address on the deployed chain.
+   * @return The address of the GHO token contract
+   */
+  function GHO_TOKEN() external view returns (address);
```
CheyenneAtapour: small nit: This ordering doesn't match the contract. `GHO_TOKEN` should come before `ROUTER` for consistency

src/bridges/ccip/AaveGhoCcipBridge.sol
```diff
@@ -0,0 +1,324 @@
...
+import {IOnRampClient} from '../../dependencies/chainlink/interfaces/IOnRampClient.sol';
+import {IRouter} from '../../dependencies/chainlink/interfaces/IRouter.sol';
+import {IRouterClient} from '../../dependencies/chainlink/interfaces/IRouterClient.sol';
+import {ITokenPool} from '../../dependencies/chainlink/interfaces/ITokenPool.sol';
+import {IAaveGhoCcipBridge} from './interfaces/IAaveGhoCcipBridge.sol';
+
+/**
+ * @title AaveGhoCcipBridge
+ * @author TokenLogic
+ * @notice It provides bridging capabilities for the GHO token across networks.
```
CheyenneAtapour: Remove the word 'It' from the beginning

```suggestion
 * @notice Provides bridging capabilities for the GHO token across networks.
```

src/bridges/ccip/interfaces/IAaveGhoCcipBridge.sol
```diff
@@ -0,0 +1,213 @@
...
+   * @notice Returns contents of specified invalid message.
+   * @param messageId The ID of the message to query
+   * @return Array of the message's tokens and amounts transferred
+   */
+  function getInvalidMessage(
+    bytes32 messageId
+  ) external view returns (Client.EVMTokenAmount[] memory);
+
+  /**
+   * @dev Returns the bridge rate limit for a given chain.
```
CheyenneAtapour:
```suggestion
   * @notice Returns the bridge rate limit for a given chain.
```

src/bridges/ccip/AaveGhoCcipBridge.sol
```diff
@@ -0,0 +1,324 @@
...
+      chainSelector,
+      amount,
+      gasLimit,
+      feeToken
+    );
+
+    uint256 fee = IRouterClient(ROUTER).getFee(chainSelector, message);
+
+    if (feeToken == address(0)) {
+      if (msg.value < fee) revert InsufficientFee();
```
CheyenneAtapour: should we refund the user ETH if they supply excess fee?

tests/bridges/ccip/AaveGhoCcipBridgeForkTest.t.sol
efecarranza: I replaced the payback in order to simplify the logic. The extra ETH can be withdrawn by the Rescuable functions and i don't really envision ETH being used much to be honest.

src/bridges/ccip/AaveGhoCcipBridge.sol
efecarranza: @CheyenneAtapour I replaced the payback in order to simplify the logic. The extra ETH can be withdrawn by the Rescuable functions and i don't really envision ETH being used much to be honest. 
The caller is always going to be governance, I'm also thinking of moving this back to Ownable.

What do you think? @miguelmtzinf

src/bridges/ccip/AaveGhoCcipBridge.sol
```diff
@@ -0,0 +1,324 @@
...
+   * @dev Checks if the destination chain has been set up and amount exceeds the rate limit
+   * @param chainSelector The chain selector of the destination chain
+   * @param amount The amount of GHO to transfer
+   */
+  function _validateDestinationAndLimit(uint64 chainSelector, uint256 amount) internal view {
+    if (destinations[chainSelector] == address(0)) {
+      revert UnsupportedChain();
+    }
+
+    uint128 limit = _getRateLimit(chainSelector);
```
miguelmtzinf, line 309: we are not checking `bridgeLimit` on ETH token pool. We can either check both things or none of them (this is for better devx, as it d revert anyway)

src/bridges/ccip/AaveGhoCcipBridge.sol
```diff
@@ -0,0 +1,324 @@
...
+      message
+    );
+
+    emit BridgeMessageInitiated(messageId, chainSelector, msg.sender, amount);
+
+    return messageId;
+  }
+
+  /// @inheritdoc CCIPReceiver
+  function ccipReceive(Client.Any2EVMMessage calldata message) external override onlyRouter {
```
efecarranza, line 113: On `processMessage()` we are checking if `message.sender` is the bridge configured.

src/bridges/ccip/AaveGhoCcipBridge.sol
efecarranza: @miguelmtzinf i have updated to check the current remaining limit (bridge - bridgedAmount),  as it's most likely going to be an unreachable statement as the rateLimit is never going to be greater than the bridgeLimit itself.

efecarranza:
@miguelmtzinf @CheyenneAtapour @brotherlymite i have updated the PR. thank you

src/bridges/ccip/AaveGhoCcipBridge.sol
```diff
@@ -0,0 +1,325 @@
...
+import {ITokenPool} from 'src/dependencies/chainlink/interfaces/ITokenPool.sol';
+import {IAaveGhoCcipBridge} from 'src/bridges/ccip/interfaces/IAaveGhoCcipBridge.sol';
+
+/**
+ * @title AaveGhoCcipBridge
+ * @author TokenLogic
+ * @notice Provides bridging capabilities for the GHO token across networks.
+ * @dev The fee token must be pre-funded for the bridge message to be sent. When doing a governance
+ * proposal ensure funds are present on this contract beforehand (GHO/LINK) or transfer funds from the
+ * treasury as part of the proposal.
```
miguelmtzinf:
```suggestion
 * @dev The fees required for bridge messages must be available in advance. For governance proposals, ensure 
 * this contract is pre-funded with the necessary tokens (GHO/LINK), or include a transfer from the treasury
 * as part of the proposal.
```

src/bridges/ccip/AaveGhoCcipBridge.sol
```diff
@@ -0,0 +1,325 @@
...
+  address public immutable GHO_TOKEN;
+
+  /// @inheritdoc IAaveGhoCcipBridge
+  address public immutable ROUTER;
+
+  /// @inheritdoc IAaveGhoCcipBridge
+  address public immutable COLLECTOR;
+
+  /// @inheritdoc IAaveGhoCcipBridge
+  mapping(uint64 chainSelector => RemoteChainConfig remoteConfig) public destinations;
```
miguelmtzinf:
```suggestion
  mapping(uint64 chainSelector => RemoteChainConfig) public destinations;
```
just because this was added in 0.8.18 and may bring a bit of friction for integrations

src/bridges/ccip/AaveGhoCcipBridge.sol
```diff
@@ -0,0 +1,325 @@
...
+ */
+contract AaveGhoCcipBridge is CCIPReceiver, Ownable, Rescuable, IAaveGhoCcipBridge {
+  using SafeERC20 for IERC20;
+
+  /// @inheritdoc IAaveGhoCcipBridge
+  bytes32 public constant BRIDGER_ROLE = keccak256('BRIDGER_ROLE');
+
+  /// https://etherscan.io/address/0x06179f7C1be40863405f374E7f5F8806c728660A
+  /// @dev Upgradeable contract so address unlikely to change
+  address public constant MAINNET_TOKEN_POOL = 0x06179f7C1be40863405f374E7f5F8806c728660A;
```
miguelmtzinf: we must fetch this from the router

src/bridges/ccip/interfaces/IAaveGhoCcipBridge.sol
```diff
@@ -0,0 +1,230 @@
...
+    bytes extraArgsOverride;
+    uint32 gasLimit;
+  }
+
+  /**
+   * @dev Emitted when a new GHO transfer is issued
+   * @param messageId The ID of the cross-chain message
+   * @param destinationChainSelector The selector of the destination chain
+   * @param from The address of sender on source chain
+   * @param amount The total amount of GHO transfered
```
miguelmtzinf:
```suggestion
   * @param amount The total amount of GHO transferred
```

src/bridges/ccip/AaveGhoCcipBridge.sol
```diff
@@ -0,0 +1,325 @@
...
+pragma solidity ^0.8.0;
+
+import {IERC165} from 'openzeppelin-contracts/contracts/utils/introspection/IERC165.sol';
+import {IERC20} from 'openzeppelin-contracts/contracts/token/ERC20/IERC20.sol';
+import {SafeERC20} from 'openzeppelin-contracts/contracts/token/ERC20/utils/SafeERC20.sol';
+import {Ownable} from 'aave-v3-origin/contracts/dependencies/openzeppelin/contracts/Ownable.sol';
+import {Rescuable} from 'solidity-utils/contracts/utils/Rescuable.sol';
+import {RescuableBase, IRescuableBase} from 'solidity-utils/contracts/utils/RescuableBase.sol';
+
+import {Client} from 'src/dependencies/chainlink/libraries/Client.sol';
```
miguelmtzinf, line 10:
```suggestion
import {RescuableBase, IRescuableBase} from 'solidity-utils/contracts/utils/RescuableBase.sol';
import {Client} from 'src/dependencies/chainlink/libraries/Client.sol';
```

src/bridges/ccip/AaveGhoCcipBridge.sol
```diff
@@ -0,0 +1,325 @@
...
+
+  /// @inheritdoc IAaveGhoCcipBridge
+  address public immutable COLLECTOR;
+
+  /// @inheritdoc IAaveGhoCcipBridge
+  mapping(uint64 chainSelector => RemoteChainConfig remoteConfig) public destinations;
+
+  /// @dev Map containing failed token transfer amounts for a message
+  mapping(bytes32 messageId => Client.EVMTokenAmount[] destTokenAmounts)
+    private _failedTokenTransfers;
```
miguelmtzinf:
```suggestion
    internal _failedTokenTransfers;
```
internal is fine, same with `_failedMessages`

src/bridges/ccip/AaveGhoCcipBridge.sol
miguelmtzinf: same with the rest

src/bridges/ccip/AaveGhoCcipBridge.sol
miguelmtzinf: sounds good. However, if the purpose of the contract is to be pre-funded always, we can remove this logic of pulling funds from sender and the payable (we should add a `receive` instead)

src/bridges/ccip/AaveGhoCcipBridge.sol
```diff
@@ -0,0 +1,325 @@
...
+
+    uint128 limit = _getRateLimit(chainSelector);
+    if (amount > limit) {
+      revert RateLimitExceeded(limit);
+    }
+
+    // Only applicable to Mainnet
+    // Should be unreachable as it's unlikely to be lower than the rate limit
+    if (block.chainid == 1) {
+      uint256 bridgeLimit = ITokenPool(MAINNET_TOKEN_POOL).getBridgeLimit() -
```
miguelmtzinf:
```suggestion
      uint256 availableToBridge = ITokenPool(MAINNET_TOKEN_POOL).getBridgeLimit() -
```

src/bridges/ccip/AaveGhoCcipBridge.sol
```diff
@@ -0,0 +1,325 @@
...
+      revert UnsupportedChain();
+    }
+
+    uint128 limit = _getRateLimit(chainSelector);
+    if (amount > limit) {
+      revert RateLimitExceeded(limit);
+    }
+
+    // Only applicable to Mainnet
+    // Should be unreachable as it's unlikely to be lower than the rate limit
```
miguelmtzinf: this is wrong, because the rate limit gets replenished based on the velocity (token per second), and bridge limit does not.

src/bridges/ccip/interfaces/IAaveGhoCcipBridge.sol
```diff
@@ -0,0 +1,230 @@
...
+  /**
+   * @notice Returns the address of the corresponding bridge for a specified chain selector.
+   * @param chainSelector The chain selector of destination chain
+   * @return The bytes representation of the address of AaveGhoCcipBridge on the destination chain
+   * @return The extra arguments to send to the destination chain in the bridge message
+   * @return The gas limit to send to the destination chain
+   */
+  function destinations(
+    uint64 chainSelector
+  ) external view returns (bytes memory, bytes memory, uint32);
```
miguelmtzinf: weird that we are exposing 3 bytes when we have our own struct for this (`RemoteChainConfig`) here, and exposing a CCIP struct (`Client.EVMTokenAmount[]`) on `getInvalidMessage`. From the integration pov i think it d be better to use our own struct for `getInvalidMessage` so it's easy to consume.

src/bridges/ccip/AaveGhoCcipBridge.sol
efecarranza: this isn't available in the router, didn't know where to fetch it from

miguelmtzinf: `router.getOnRamp(destionationChainSelector).getPoolBySourceToken(destionationChainSelector, GHO_TOKEN)`

src/bridges/ccip/AaveGhoCcipBridge.sol
```diff
@@ -0,0 +1,331 @@
...
+    address initialOwner
+  ) CCIPReceiver(router) {
+    ROUTER = router;
+    GHO_TOKEN = gho;
+    COLLECTOR = collector;
+
+    transferOwnership(initialOwner);
+  }
+
+  /// @dev Default receive function so the contract can be sent Ether
```
miguelmtzinf:
```suggestion
  /// @dev Default receive function enabling the contract to accept native tokens
```

src/bridges/ccip/AaveGhoCcipBridge.sol
```diff
@@ -0,0 +1,331 @@
...
+  address public immutable GHO_TOKEN;
+
+  /// @inheritdoc IAaveGhoCcipBridge
+  address public immutable ROUTER;
+
+  /// @inheritdoc IAaveGhoCcipBridge
+  address public immutable COLLECTOR;
+
+  /// @dev Map containing destination chain's configuration
+  mapping(uint64 => RemoteChainConfig) internal destinations;
```
miguelmtzinf:
```suggestion
  mapping(uint64 => RemoteChainConfig) internal _destinations;
```

src/bridges/ccip/AaveGhoCcipBridge.sol
```diff
@@ -0,0 +1,331 @@
...
+      revert InvalidZeroAddress();
+    }
+
+    destinations[chainSelector] = RemoteChainConfig({
+      destination: destination,
+      extraArgsOverride: extraArgs,
+      gasLimit: gasLimit
+    });
+
+    emit DestinationChainSet(chainSelector, destination, gasLimit);
```
miguelmtzinf: not emitting `extraArgs` here.

src/bridges/ccip/AaveGhoCcipBridge.sol
```diff
@@ -0,0 +1,331 @@
...
+    RemoteChainConfig memory remoteConfig = destinations[chainSelector];
+
+    Client.EVM2AnyMessage memory message = Client.EVM2AnyMessage({
+      receiver: remoteConfig.destination,
+      data: '',
+      tokenAmounts: tokenAmounts,
+      extraArgs: remoteConfig.extraArgsOverride.length > 0
+        ? remoteConfig.extraArgsOverride
+        : Client._argsToBytes(
+          Client.GenericExtraArgs({gasLimit: remoteConfig.gasLimit, allowOutOfOrderExecution: true})
```
miguelmtzinf, line 276: so how `gasLimit` would work here? it will always use the configured value and never the default?

efecarranza: Yes, it will use the configured, I ran tests with chainlink and not using the default for EVM chains would save > 100,000 gas.

efecarranza: I added to the README a description on gas to help set it

https://github.com/bgd-labs/aave-helpers/pull/443/files#diff-cf147313cc303a0f6ebf3b1c64e67a2ec7295cabba74bec81e535b519aa15857R40

src/bridges/ccip/README.md
```diff
@@ -0,0 +1,46 @@
...
+# Destination Chains
+
+The chain selector for a given chain is defined by CCIP. The list of chain selectors can be found here:
+https://docs.chain.link/ccip/directory/mainnet
+
+Non-EVM chains are supported.
+
+### Destination Addresses
+
+The destination chain address must be provided as bytes. For EVM chains, convert the address to bytes with `abi.encode(address)`, where address if an EVM address. For Solana/Aptos, `abi.encode(bytes32)` where bytes32 is the Aptos/Solana address.
```
CheyenneAtapour: We can remove this paragraph, favoring the one below

src/bridges/ccip/interfaces/IAaveGhoCcipBridge.sol
```diff
@@ -0,0 +1,233 @@
...
+/**
+ * @title IAaveGhoCcipBridge
+ * @author TokenLogic
+ * @notice Defines the behaviour of an AaveGhoCcipBridge
+ */
+interface IAaveGhoCcipBridge {
+  /**
+   * Struct representing a destination chain
+   * @return extraArgsOverride Any extra arguments to pass with message to the destination chain
+   * @return bridge The bytes representation of an address (EVM or non-EVM)
```
CheyenneAtapour: this should be `destination` and should be the first return

src/dependencies/chainlink/interfaces/IRouter.sol
```diff
@@ -0,0 +1,38 @@
...
+
+  /// @notice Route the message to its intended receiver contract.
+  /// @param message Client.Any2EVMMessage struct.
+  /// @param gasForCallExactCheck of params for exec.
+  /// @param gasLimit set of params for exec.
+  /// @param receiver set of params for exec.
+  /// @dev if the receiver is a contracts that signals support for CCIP execution through EIP-165.
+  /// the contract is called. If not, only tokens are transferred.
+  /// @return success A boolean value indicating whether the ccip message was received without errors.
+  /// @return retBytes A bytes array containing return data form CCIP receiver.
```
CheyenneAtapour:
```suggestion
  /// @return retBytes A bytes array containing return data from CCIP receiver.
```

src/dependencies/chainlink/interfaces/IRouterClient.sol
```diff
@@ -0,0 +1,37 @@
...
+  error InvalidMsgValue();
+
+  /// @notice Checks if the given chain ID is supported for sending/receiving.
+  /// @param destChainSelector The chain to check.
+  /// @return supported is true if it is supported, false if not.
+  function isChainSupported(uint64 destChainSelector) external view returns (bool supported);
+
+  /// @param destinationChainSelector The destination chainSelector.
+  /// @param message The cross-chain CCIP message including data and/or tokens.
+  /// @return fee returns execution fee for the message.
```
CheyenneAtapour, line 18:
```suggestion
  /// @return fee returns execution fee for the message
```
Looks like the sentence is meant to be continued below

src/dependencies/chainlink/interfaces/IRouterClient.sol
```diff
@@ -0,0 +1,37 @@
...
+  function getFee(
+    uint64 destinationChainSelector,
+    Client.EVM2AnyMessage memory message
+  ) external view returns (uint256 fee);
+
+  /// @notice Request a message to be sent to the destination chain.
+  /// @param destinationChainSelector The destination chain ID.
+  /// @param message The cross-chain CCIP message including data and/or tokens.
+  /// @return messageId The message ID.
+  /// @dev Note if msg.value is larger than the required fee (from getFee) we accept.
```
CheyenneAtapour:
```suggestion
  /// @dev Note if msg.value is larger than the required fee (from getFee) we accept
```
Same, sentence continued in next line

CheyenneAtapour (commented):
Looks good to us

-Aave Labs

src/dependencies/chainlink/interfaces/IRouterClient.sol
CheyenneAtapour: Ah, sry, this is a chainlink dep. So all good!

brotherlymite (approved):

PR #469, opened 2025-07-30
feat: whitelabel

brotherlymite:


src/GovV3Helpers.sol
```diff
@@ -394,6 +426,12 @@ library GovV3Helpers {
     payloadsController.executePayload(readyPayload(vm, payloadAddress));
   }
 
+  function executePayload(Vm vm, address payloadAddress, address payloadsController) internal {
```
ianflexa, line 435: this will be used for the custom/white-label payloadsController, right? 
Can you add natspec docs

src/GovV3Helpers.sol
```diff
@@ -412,6 +450,24 @@ library GovV3Helpers {
     return payloadId;
   }
 
+  function readyPayload(
```
ianflexa, line 459: same here

pavelvm5 (approved):

ianflexa (approved):

PR #547, opened 2025-11-06
feat: update tooling to support multi-chain deploy

brotherlymite:
Adds a method `deployRegisterPayload()` to deploy and register payload on the given network. 
Used in https://github.com/bgd-labs/aave-proposals-v3/pull/890 to allow multi chain deployment for proposals.

### DIFF
--- a/src/GovV3Helpers.sol
+++ b/src/GovV3Helpers.sol
@@ -262,6 +262,25 @@ library GovV3Helpers {
   }
 
   /**
+   * @notice method to deploy the payload contracts and register them on payloads controller for a given network
+   * @param chainId the chain id for the network where payload will be deployed and registered
+   * @param payloadBytecode list of payload creation bytecode which will be deployed and registered
+   */
+  function deployRegisterPayload(Vm vm, uint256 chainId, bytes[] memory payloadBytecode) internal {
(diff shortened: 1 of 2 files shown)


sakulstra (approved):

src/GovV3Helpers.sol
```diff
@@ -261,6 +261,25 @@ library GovV3Helpers {
     return bytecodeHash;
   }
 
+  /**
+   * @notice method to deploy the payload contracts and register them on payloads controller for a given network
+   * @param chainId the chain id for the network where payload will be deployed and registered
+   * @param payloadBytecode list of payload creation bytecode which will be deployed and registered
+   */
+  function deployRegisterPayload(Vm vm, uint256 chainId, bytes[] memory payloadBytecode) internal {
+    ChainHelpers.selectChain(vm, chainId);
```
sakulstra, line 270: is it fine / correct to "stay in the chain", after the script ends?

brotherlymite: I think after script ends we can select the previous chain which was selected. 

The thing is on proposals the chain will be selected initially from:
- `Anvil -> ChainX`

Then I think it is not so trivial to go from `ChainX -> Anvil` via `createSelectFork()` on forge

PR #676, opened 2026-03-24
fix: use getStateDiffJson instead of getStorageAccesses to avoid MemoryOOG

NandyBa:
## Summary

- Replaces `vm.getStorageAccesses()` with `vm.getStateDiffJson()` in `_validateNoExecutorStorageChange` to avoid `MemoryOOG` errors on large proposals
- Reorders operations in `defaultTest` so raw diff and logs are captured before `createConfigurationSnapshot`, matching the execution flow that was causing the issue
- Updates `ProtocolV2TestBase` with the same fix pattern

## Motivation

Proposals that touch a large number of reserves (e.g. multi-asset listings) were triggering a `MemoryOOG` error because `vm.getStorageAccesses()` materializes all storage accesses into a Solidity array in memory. Switching to `vm.getStateDiffJson()` serializes the diff to JSON without the intermediate in-memory array, avoiding the OOG.

### Reproducer

- **Affected branch (MemoryOOG):** https://github.com/AaveChan/aave-proposals-v3/tree/srUSDe-June-listing-on-core
- **Fixed branch (no MemoryOOG):** https://github.com/AaveChan/aave-proposals-v3/tree/srUSDe-June-listing-on-core-oog-fix
- **Diff between the two:** https://github.com/AaveChan/aave-proposals-v3/pull/32

The only difference between both branches is the workaround applied in the proposal test to avoid the OOG — this fix makes that workaround unnecessary.

**Notice that due to some use database, I was not able to pass these tests**


sakulstra (approved):
ver clever 😅

PR #690, opened 2026-04-09
test: aave v4 e2e

yan-man:
## Description

- port ProtocolV4TestBase for use during AIP execution
- add tests for ProtocolV4TestBase
- add dependencies for v4-specific contracts.

## Checklist

- [ ] Added a [changeset](https://github.com/changesets/changesets) if this modifies a published package (`pnpm changeset`)


foundry.toml
```diff
@@ -5,9 +5,10 @@ script = 'scripts'
 out = 'out'
 libs = ['lib']
 remappings = []
-fs_permissions = [{ access = "read-write", path = "./reports" }]
+fs_permissions = [{ access = "read-write", path = "./reports" }, { access = "read-write", path = "./diffs" }]
 ffi = true
 evm_version = 'cancun'
+gas_limit = 9223372036854775807
```
DhairyaSethi: this shouldn't be needed if we do the other more correct fix which is pause gas metering before running e2e tests and resume right after no?

foundry.toml
```diff
@@ -5,9 +5,10 @@ script = 'scripts'
 out = 'out'
 libs = ['lib']
 remappings = []
-fs_permissions = [{ access = "read-write", path = "./reports" }]
+fs_permissions = [{ access = "read-write", path = "./reports" }, { access = "read-write", path = "./diffs" }]
```
DhairyaSethi, line 8: we did this bc of time but now the snapshot diff to md logic can be moved to aave-helpers-js, like v3, instead of being in solidity. it has some reusability benefits bc it's used in seatbelt also

DhairyaSethi: and then the e2e test can just ffi

src/ProtocolV4TestBase.sol
```diff
@@ -0,0 +1,776 @@
...
+    address payload
+  ) private returns (string memory rawDiff, string memory logsJson) {
+    uint256 startGas = gasleft();
+    vm.startStateDiffRecording();
+    vm.recordLogs();
+
+    GovV3Helpers.executePayload(
+      vm,
+      payload,
+      address(GovV3Helpers.getPayloadsController(ChainIds.MAINNET))
```
DhairyaSethi: should be chain id agnostic no?

src/ProtocolV4TestBase.sol
```diff
@@ -0,0 +1,776 @@
...
+    logsJson = vm.getRecordedLogsJson();
+  }
+
+  /// @notice Test all reserves on every spoke in the array.
+  function e2eTestAllSpokes(ISpoke[] memory spokes) public {
+    for (uint256 i; i < spokes.length; i++) {
+      console.log('--- E2E: Testing spoke %s ---', address(spokes[i]));
+      console.log('--------------------------------');
+      e2eTestSpoke(spokes[i]);
+      e2eTestPositionManagers(spokes[i]);
```
DhairyaSethi: i think we can avoid testing posms here or atleast put it behind another flag

foundry.toml
yan-man: hm ok can try this

foundry.toml
yan-man: I'll do this in a follow up pr

src/ProtocolV4TestBase.sol
yan-man: https://github.com/aave-dao/aave-helpers/pull/690/changes/0dc6a717c7d6a2be63a4cf95c80b6458451ee1e3

src/ProtocolV4TestBase.sol
yan-man: https://github.com/aave-dao/aave-helpers/pull/690/changes/0dc6a717c7d6a2be63a4cf95c80b6458451ee1e3

foundry.toml
yan-man: https://github.com/aave-dao/aave-helpers/pull/690/changes/0dc6a717c7d6a2be63a4cf95c80b6458451ee1e3

Kogaroshi:
Could you also add a `src/v4-config-engine/AaveV4PayloadEthereum.sol` so we have all the base required for proposals-v3 ?

yan-man:
refer to https://github.com/aave-dao/aave-helpers/pull/694 instead to help pick up megaeth RPC URL properly

PR #694, opened 2026-04-16
feat: Aave V4 e2e tests

yan-man:
## Description

- Aave V4 e2e tests built in the same style as for Aave V3
- accompanying local v4 dependencies and helpers
- AaveV4PayloadEthereum

## Checklist

- [x] Added a [changeset](https://github.com/changesets/changesets) if this modifies a published package (`pnpm changeset`)


foundry.toml
```diff
@@ -5,7 +5,7 @@ script = 'scripts'
 out = 'out'
 libs = ['lib']
 remappings = []
-fs_permissions = [{ access = "read-write", path = "./reports" }]
+fs_permissions = [{ access = "read-write", path = "./reports" }, { access = "read-write", path = "./diffs" }]
```
DhairyaSethi: let's do json and md generation in ts rather than solidity ser

yan-man: i was mentioning from other pr, just want to do this as a follow up pr so we can get the first version in rn

yan-man: or do you think it's needed here in this first version too?

DhairyaSethi: work with claude i think it shouldn't be good big an effort, original solidity one is also mostly ai. it needs to be in js package to be reused by downstream

src/dependencies/v4/AaveV4EthereumHelpers.sol
```diff
@@ -0,0 +1,78 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import {IHub, ISpoke, ITokenizationSpoke} from 'aave-address-book/AaveV4.sol';
+import {AaveV4EthereumHubs, AaveV4EthereumSpokes, AaveV4EthereumTokenizationSpokes} from 'aave-address-book/AaveV4Ethereum.sol';
+
+/// @dev Helper that returns all hubs as an array (address-book only exposes individual constants).
+library AaveV4EthereumHubHelpers {
+  function getHubs() internal pure returns (IHub[] memory) {
+    IHub[] memory hubs = new IHub[](3);
```
DhairyaSethi, line 10: i think address book should be able to create this dynamically from the config, otherwise rn we need to update this manually as well which is super annoying

src/dependencies/v4/Actions.sol
```diff
@@ -0,0 +1,467 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import 'forge-std/Test.sol';
+import {IERC20} from 'openzeppelin-contracts/contracts/token/ERC20/IERC20.sol';
+import {SafeERC20} from 'openzeppelin-contracts/contracts/token/ERC20/utils/SafeERC20.sol';
```
DhairyaSethi:
```suggestion
import {SafeERC20, IERC20} from 'openzeppelin-contracts/contracts/token/ERC20/utils/SafeERC20.sol';
```

also import order below, interfaces then types

src/dependencies/v4/Actions.sol
```diff
@@ -0,0 +1,467 @@
...
+    assertGe(
+      snapshotAfter.spokeOnHub.drawnDebt,
+      snapshotBefore.spokeOnHub.drawnDebt,
+      'TIME_SKIP: hub spoke drawn decreased'
+    );
+
+    // Hub drawn index should have grown
+    IHubBase hub = IHubBase(reserveInfo.hub);
+    uint256 drawnIndexAfter = hub.getAssetDrawnIndex(reserveInfo.assetId);
+    assertGt(drawnIndexAfter, 1e27, 'TIME_SKIP: drawn index should be greater than 1e27');
```
DhairyaSethi: use ray instead since v4 available? nbd

src/dependencies/v4/Actions.sol
```diff
@@ -0,0 +1,467 @@
...
+  }
+
+  function _supply(
+    ISpoke spoke,
+    Types.ReserveInfo memory reserveInfo,
+    address user,
+    uint256 amount
+  ) internal {
+    require(!reserveInfo.paused, 'SUPPLY: PAUSED_RESERVE');
+    require(!reserveInfo.frozen, 'SUPPLY: FROZEN_RESERVE');
```
DhairyaSethi, line 170: should this require? dont we have an issue on v3 tests also rn where we weren't able to test freeze of scroll bc of this?

not sure of the best approach tho, if we should test frozen cases (ie should be able to supply) or skip

src/dependencies/v4/Actions.sol
```diff
@@ -0,0 +1,467 @@
...
+
+    Types.PositionSnapshot memory snapshotBefore = _getPositionSnapshot(spoke, reserveInfo, user);
+
+    vm.startPrank(user);
+    deal2(reserveInfo.underlying, user, amount);
+    IERC20(reserveInfo.underlying).forceApprove(address(spoke), amount);
```
DhairyaSethi: this can be abstracted in an internal helper, its reused a bit and can be useful while testing

src/dependencies/v4/Actions.sol
```diff
@@ -0,0 +1,467 @@
...
+      snapshotAfter.spokeOnHub.totalDebt,
+      snapshotBefore.spokeOnHub.totalDebt + amount,
+      2,
+      'BORROW: hub debt mismatch'
+    );
+    assertEq(
+      snapshotAfter.spokeOnHub.drawnShares,
+      snapshotBefore.spokeOnHub.drawnShares + expectedDrawnShares,
+      'BORROW: hub drawn shares mismatch'
+    );
```
DhairyaSethi, line 328: should we also specifically test any risk premium cases - incase config gets fked up? nbd now and unsure how we will, lets start a ticket w ideas

src/dependencies/v4/Actions.sol
```diff
@@ -0,0 +1,467 @@
...
+    address user,
+    uint256 amount
+  ) internal {
+    Types.PositionSnapshot memory snapshotBefore = _getPositionSnapshot(spoke, reserveInfo, user);
+    uint256 effectiveRepayAmount = amount >= snapshotBefore.user.totalDebt
+      ? snapshotBefore.user.totalDebt
+      : amount;
+    uint256 drawnRepayAmount = effectiveRepayAmount > snapshotBefore.user.drawnDebt
+      ? snapshotBefore.user.drawnDebt
+      : effectiveRepayAmount;
```
DhairyaSethi: this logic is incorrect, premium debt is restored first

src/dependencies/v4/Actions.sol
```diff
@@ -0,0 +1,467 @@
...
+    uint256 expectedRestoredShares = IHubBase(reserveInfo.hub).previewRestoreByAssets(
+      reserveInfo.assetId,
+      drawnRepayAmount
+    );
+
+    vm.startPrank(user);
+    // deal enough to cover full repay, capped to avoid overflow
+    uint256 maxDeal = _maxDealAmount(reserveInfo.decimals);
+    deal2(reserveInfo.underlying, user, maxDeal);
```
DhairyaSethi: i mean you've calced effective repay amount so should just use that no?  dont need to worry about uint max overflow

src/dependencies/v4/AaveV4EthereumHelpers.sol
yan-man: ok we can fix this upstream in address book instead, i'll leave this here for now until we have it there?

src/dependencies/v4/Actions.sol
```diff
@@ -0,0 +1,467 @@
...
+    );
+    Types.PositionSnapshot memory debtSnapshotBefore = _getPositionSnapshot(
+      spoke,
+      debtInfo,
+      borrower
+    );
+    assertGt(debtSnapshotBefore.user.totalDebt, 0, 'LIQUIDATE: borrower has no debt');
+
+    vm.startPrank(liquidator);
+    uint256 dealAmount = _maxDealAmount(debtInfo.decimals);
```
DhairyaSethi: just cap it to total debt no

src/dependencies/v4/Actions.sol
yan-man: but this is a supply action so if it is frozen then we cannot supply here no? i think freeze testing shouldnt exist on e2e test as (imo) it is meant to be happy path flows for end user

src/dependencies/v4/Actions.sol
```diff
@@ -0,0 +1,467 @@
...
+      assertApproxEqAbs(
+        stdMath.delta(snapshotAfter.user.totalDebt, snapshotBefore.user.totalDebt),
+        amount,
+        2,
+        'REPAY: user debt mismatch'
+      );
+    }
+    // Hub spoke - up to 2 wei diff due to premium/drawn debt
+    assertApproxEqAbs(
+      stdMath.delta(snapshotBefore.spokeOnHub.totalDebt, snapshotAfter.spokeOnHub.totalDebt),
```
DhairyaSethi: can do using syntax

src/dependencies/v4/Actions.sol
DhairyaSethi: also can we add one more here and after withdraw that hf > 1

src/dependencies/v4/GatewayScenarios.sol
```diff
@@ -0,0 +1,1012 @@
...
+  ) internal view returns (bytes memory) {
+    bytes32 digest = keccak256(
+      abi.encodePacked('\x19\x01', gateway.DOMAIN_SEPARATOR(), structHash)
+    );
+    (uint8 v, bytes32 r, bytes32 s) = vm.sign(privateKey, digest);
+    return abi.encodePacked(r, s, v);
+  }
+
+  // -------------------------------------------------------------------------
+  // NativeTokenGateway scenario
```
DhairyaSethi, line 52: i think we can avoid testing these tbh

src/dependencies/v4/Helpers.sol
```diff
@@ -0,0 +1,393 @@
...
+
+/// @title Helpers
+/// @notice Query and utility functions for V4 e2e tests.
+abstract contract Helpers is Actions {
+  /// @notice Build ReserveInfo[] for all reserves on a spoke.
+  function _getReserveInfo(ISpoke spoke) internal view returns (Types.ReserveInfo[] memory) {
+    uint256 count = spoke.getReserveCount();
+    Types.ReserveInfo[] memory info = new Types.ReserveInfo[](count);
+
+    for (uint256 i; i < count; i++) {
```
DhairyaSethi:
```suggestion
    for (uint256 reserveId; reserveId < count; ++reserveId) {
```

src/dependencies/v4/Helpers.sol
```diff
@@ -0,0 +1,393 @@
...
+      ISpoke.Reserve memory reserve = spoke.getReserve(i);
```
DhairyaSethi: dont need to cache these three here, just call them while building the struct

src/dependencies/v4/Helpers.sol
```diff
@@ -0,0 +1,393 @@
...
+
+  /// @notice Supply collateral to borrower on the same spoke, then enable as collateral.
+  ///         Assumes addCaps have been set to max via _setCapsToMax before calling.
+  function _ensureCollateral(
+    ISpoke spoke,
+    Types.ReserveInfo memory reserveInfo,
+    address borrower,
+    uint256 amount
+  ) internal {
+    _supply({spoke: spoke, reserveInfo: reserveInfo, user: borrower, amount: amount});
```
DhairyaSethi: named args can be avoided here

src/dependencies/v4/Helpers.sol
```diff
@@ -0,0 +1,393 @@
...
+  ///         Compares against CF-adjusted totalCollateralValue, so it may use multiple reserves.
+  function _ensureBorrowCapacity(
+    ISpoke spoke,
+    address borrower,
+    uint256 borrowAmountInDollars
+  ) internal {
+    Types.ReserveInfo[] memory goodCollaterals = _getAllUsableCollaterals(_getReserveInfo(spoke));
+    address oracleAddr = spoke.ORACLE();
+    uint8 oracleDecimals = IAaveOracle(oracleAddr).decimals();
+    uint256 targetCollateralDollarAmount = borrowAmountInDollars * 3;
```
DhairyaSethi, line 126: instead of InDollars maybe use Value? (not exactly the same bc value is wadified $$ but yk)

src/dependencies/v4/Helpers.sol
```diff
@@ -0,0 +1,393 @@
...
+        addCap: type(uint40).max
+      });
+    }
+    vm.clearMockedCalls();
+  }
+
+  /// @notice Safely get the ERC20 symbol, fallback to "UNKNOWN".
+  function _safeSymbol(address token) internal view returns (string memory) {
+    try IERC20Metadata(token).symbol() returns (string memory s) {
+      return s;
```
DhairyaSethi: symbol should always be implemented, its part of the standard (decimals isn't)

this can be removed

src/dependencies/v4/Helpers.sol
```diff
@@ -0,0 +1,393 @@
...
+    } catch {
+      return 'UNKNOWN';
+    }
+  }
+
+  /// @notice Half a token in the asset's native decimals.
```
DhairyaSethi: kill weird helper pls lol

src/dependencies/v4/GatewayScenarios.sol
yan-man: i have these tests in here but have set the `testPositionManagers: false` in defaultTest so they're not tested

src/dependencies/v4/Helpers.sol
yan-man: eh i think that's kind of repetitive, prefer to just cache it here first

src/dependencies/v4/Helpers.sol
yan-man: Prefer to just use dollars for clarity, vs needing to define "value" somewhere

src/dependencies/v4/Helpers.sol
yan-man: lmao created this before when we only had WBTC cap at 1 unit

src/dependencies/v4/AaveV4EthereumHelpers.sol
DhairyaSethi: yeah pls create a ticket somewhere

src/dependencies/v4/Actions.sol
DhairyaSethi: yeah for supply it should skip but wdyt in general? curious about @avniculae thoughts on this

src/dependencies/v4/Scenarios.sol
```diff
@@ -0,0 +1,690 @@
...
+    // simple check - ensure at least 1 share worth of debt assets is liquidated
+    // technically possible to liquidate less if premium debt exists, but serves as a basic check
+    if (
+      spoke.getReserve(testAssetInfo.reserveId).hub.previewRestoreByAssets(
+        testAssetInfo.reserveId,
+        partialDebt
+      ) == 0
+    ) {
+      partialDebt = spoke.getReserve(testAssetInfo.reserveId).hub.previewRestoreByShares(
+        testAssetInfo.reserveId,
```
DhairyaSethi: hub takes assetId, not reserveId

kinda scary tests pass and wasn't able to catch this?

src/dependencies/v4/Scenarios.sol
```diff
@@ -0,0 +1,690 @@
...
+      liquidator: liquidator,
+      borrower: borrower,
+      debtToCover: partialDebt,
+      receiveShares: receiveShares
+    });
+    assertGt(
+      spoke.getUserTotalDebt(testAssetInfo.reserveId, borrower),
+      0,
+      'PARTIAL_LIQUIDATION: debt should not be fully repaid'
+    );
```
DhairyaSethi, line 457: maybe we can add more post liq assertions? one im thinking of is liquidation was profitable to liquidator to check params weren't fked

src/dependencies/v4/Scenarios.sol
yan-man: this was a last min addition to just catch a partial liq that was equal to 0, so it would just skip here given it was reserveId instead of assetId, but will fix

src/dependencies/v4/Actions.sol
yan-man: on the RP what are you thinking, like you mean verifying user RP is as expected? yeah will leave it for now and open a linear ticket on it

Hf validation has been added

src/dependencies/v4/Scenarios.sol
yan-man: added some more, see wdyt

foundry.toml
yan-man: added js based one

packages/aave-helpers-js/package.json
DhairyaSethi, line 3: do we need to inc version or does ci release please take care of this?

packages/aave-helpers-js/package.json
```diff
@@ -5,9 +5,9 @@
   "type": "module",
   "exports": {
     ".": {
+      "types": "./dist/index.d.ts",
       "import": "./dist/index.mjs",
-      "require": "./dist/index.cjs",
-      "types": "./dist/index.d.ts"
+      "require": "./dist/index.cjs"
```
DhairyaSethi: why? revert pls

yan-man: i was getting warnings during running js tests, reverted

packages/aave-helpers-js/package.json
yan-man: added a changeset to take care of this, wdyt 1.1.0?

src/ProtocolV4TestBase.sol
```diff
@@ -0,0 +1,827 @@
...
+/// Tests NativeTokenGateway and SignatureGateway for each spoke.
+/// Loops over all good collaterals and uses randomized amounts.
+contract ProtocolV4TestBase is SnapshotV4, Scenarios, TokenizationScenarios, GatewayScenarios {
+  using SafeERC20 for IERC20;
+
+  /// @notice Run the full V4 test suite: snapshot before, execute payload, snapshot after, diff, then e2e.
+  function defaultTest(
+    string memory reportName,
+    ISpoke[] memory spokes,
+    address[] memory tokenizationSpokes,
```
Kogaroshi, line 33: future update : we could use getter to get all Spokes & all TokenizationSpokes for a given network, and just have `defaultTest(reportName, chainId, payload)` (or other way to get the specific network list of spokes) for simpler integration in Proposals

src/ProtocolV4TestBase.sol
```diff
@@ -0,0 +1,827 @@
...
+        spoke: spoke,
+        reserveInfo: goodDebtReserves[0],
+        collateralInfo: goodCollaterals[0]
+      });
+      vm.revertToState(gatewaySnapshot);
+    }
+  }
+
+  /// @notice Test that a frozen reserve correctly reverts on supply and borrow.
+  function e2eTestFrozenAsset(ISpoke spoke, Types.ReserveInfo memory frozenAsset) public {
```
Kogaroshi: maybe move with e2eTestAsset & the pause variant ?

Kogaroshi (approved):

src/ProtocolV4TestBase.sol
yan-man: so do you mean calling the getter from within the defaultTest so it doesnt need to be passed into this fn?

Kogaroshi: yup

Kogaroshi (approved):

src/dependencies/v4/AaveV4EthereumHelpers.sol
yan-man: ticket has been created in linear

src/dependencies/v4/Actions.sol
avniculae: I think it makes sense to keep as is for now, we can think about how to deal with this more holistically

src/dependencies/v4/Actions.sol
avniculae: can we open a github issue as well?

yan-man: added to the other linear ticket re: here https://github.com/aave-dao/aave-helpers/pull/694#discussion_r3209744835

src/dependencies/v4/SnapshotV4.sol
```diff
@@ -0,0 +1,228 @@
...
+      './diffs/',
+      reportName,
+      '_before_',
+      reportName,
+      '_after.md'
+    );
+
+    string[] memory inputs = new string[](7);
+    inputs[0] = 'npx';
+    inputs[1] = '@aave-dao/aave-helpers-js@^1.0.1';
```
avniculae, line 44: should we pin maybe? or not?

yan-man: do you mean pin to 1.1.0? (given the new version which will be published)

avniculae: I see this is pattern is used for v3 as well, but maybe we cna do `^1.1.0` since v4 is not available before, nbd I guess

avniculae (approved):

PR #743, opened 2026-06-26
fix: measure payload gas against real per-chain tx limit under isolation

avniculae:
## Problem

The payload gas-limit check in `ProtocolV3TestBase`/`ProtocolV4TestBase` was effectively a no-op and, where it did run, under-counted gas:

- It compared `gasleft()` usage against **`block.gaslimit`**, which foundry overrides to a huge value (e.g. `9223372036854775807` in the proposals `test` profile). `95%` of that is never exceeded, so the assertion never fired.
- It measured the payload **without isolation**, so execution ran against warm storage and reported far less than a real (cold) mainnet transaction.

## Fix

- **Real limit.** Replace `block.gaslimit` with `_getMaxPayloadGas()`, returning EIP-7825's per-transaction gas cap (`16_777_216`). It's applied as a conservative floor for every chain — a payload that fits within it is executable on any supported chain (no listed chain currently has a lower per-tx cap) — and is `virtual` so a specific chain/test can raise it where a higher limit is genuinely needed.
- **Accurate cold metering.** Measure the payload on a dedicated isolated run via `vm.lastCallGas().gasTotalUsed`, taken inside a `snapshotState`/`revertToState` so it has no effect on the recorded state diff. `vm.startStateDiffRecording()` (required for the diff + executor check) suppresses isolate's transaction-level gas accounting, so the metered run keeps recording **off** and the report is produced on a separate recorded run. See https://github.com/foundry-rs/foundry/issues/15381
- **Isolation.** Enable isolation on the gas-path tests via the inline `/// forge-config: default.isolate = true` annotation (per-test, no global config change).

## Validation

Verified against a real, near-limit payload (V4 onboarding, executed on mainnet at `15,657,968` gas per the on-chain transaction): the check now reports the gas **bit-for-bit** (`15,657,968`), vs the old check which passed it trivially against the overridden `block.gaslimit`.

### Why a single floor instead of per-chain values

The per-transaction cap is a consensus/admission rule (EIP-7825), not something observable from a fork or a read RPC: `block.gaslimit` is overridden by foundry, there is no RPC method for it, and simulation (`eth_call` / `eth_estimateGas` / `tenderly_simulateTransaction`) does not enforce it — only real `eth_sendRawTransaction` does. Hardcoding per-chain block limits is fragile (they drift, and several chains use dynamic limits), so we use the strictest deterministic cap (EIP-7825's `2^24`) as a floor for all chains, with the `virtual` override as the escape hatch.

### DIFF
--- a/src/CommonTestBase.sol
+++ b/src/CommonTestBase.sol
@@ -50,8 +50,18 @@ contract CommonTestBase is Test {
     return 16_777_216;
   }
 
-  function _assertPayloadGasWithinLimit(uint256 gasUsed) internal view {
+  function _assertPayloadGasWithinLimit(uint256 gasUsed) internal {
+    _requireIsolation();
     assertLt(gasUsed, (_getMaxPayloadGas() * 95) / 100, 'TX_GAS_LIMIT_EXCEEDED'); // 5% is kept as a buffer
+  }
+
+  // The gas check is only meaningful under isolation (cold storage + tx intrinsic), and there is no
+  // cheatcode to query or enable it. Detect it instead: under isolation a top-level call is metered as
+  // a transaction (>= 21000 intrinsic gas), otherwise ~0. Fail loudly rather than silently under-count.
+  function _requireIsolation() internal {
+    (bool success, ) = address(0).call('');
+    success;
+    require(vm.lastCallGas().gasTotalUsed >= 21_000, 'PAYLOAD_GAS_CHECK_REQUIRES_ISOLATION');
   }
 
   /**
--- a/src/CommonTestBase.sol
+++ b/src/CommonTestBase.sol
@@ -43,6 +43,17 @@ contract CommonTestBase is Test {
     );
   }
 
+  // EIP-7825 (Fusaka) per-tx gas cap, used as a conservative floor for every chain: a payload that fits
+  // within it is executable on any supported chain. Can be replaced with an on-chain read once EIP-8123
+  // is accepted, or overridden per chain where a higher limit is needed.
+  function _getMaxPayloadGas() internal view virtual returns (uint256) {
+    return 16_777_216;
+  }
+
+  function _assertPayloadGasWithinLimit(uint256 gasUsed) internal view {
+    assertLt(gasUsed, (_getMaxPayloadGas() * 95) / 100, 'TX_GAS_LIMIT_EXCEEDED'); // 5% is kept as a buffer
+  }
+
   /**
    * @notice deal doesn't support amounts stored in a script right now.
    * This function patches deal to mock and transfer funds instead.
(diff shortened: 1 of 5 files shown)


tests/ProtocolV3TestBase.t.sol
```diff
@@ -157,6 +157,7 @@ contract ProtocolV3TestMegaEthSnapshot is ProtocolV3TestBase {
     vm.createSelectFork('megaeth', 7862955);
   }
 
+  /// forge-config: default.isolate = true
   function test_snapshotState() public {
```
DhairyaSethi, line 161: we will need to make the proposals tests isolate true bc otherwise it wouldnt work, are there cheatcodes available now where we can assert this mode is on at runtime or set it without comments which is enforced always

avniculae: [b8719e0](https://github.com/aave-dao/aave-helpers/pull/743/commits/b8719e07bc26eb49672ff2339eccf67bc06cccd8)

src/CommonTestBase.sol
```diff
@@ -43,6 +43,27 @@ contract CommonTestBase is Test {
...
+    assertLt(gasUsed, (_getMaxPayloadGas() * 95) / 100, 'TX_GAS_LIMIT_EXCEEDED'); // 5% is kept as a buffer
+  }
+
+  // The gas check is only meaningful under isolation (cold storage + tx intrinsic), and there is no
+  // cheatcode to query or enable it. Detect it instead: under isolation a top-level call is metered as
+  // a transaction (>= 21000 intrinsic gas), otherwise ~0. Fail loudly rather than silently under-count.
+  function _requireIsolation() internal {
+    (bool success, ) = address(0).call('');
+    success;
+    require(vm.lastCallGas().gasTotalUsed >= 21_000, 'PAYLOAD_GAS_CHECK_REQUIRES_ISOLATION');
```
DhairyaSethi, line 64: let's give the reader a hint on how to fix this as well saying `isolate = true` exactly

avniculae: we have ` * @dev the calling test must run under isolation` on top of defaultTest. you want something else?

DhairyaSethi: `under isolation` is not clear it refers to isolation mode config, i would be very explicit by saying

```
 add a comment above your function which is `// forge-config profile.default.isolate = true`
 ```

avniculae: [77e7a55](https://github.com/aave-dao/aave-helpers/pull/743/commits/77e7a55d1b59e1b71286fbc2d6d98d94b5fea351)

DhairyaSethi: we will have to revisit incase gas repricing on eth_call base cost but i think this is a fine assumption to make for now

DhairyaSethi: another dynamic idea would be measure gas cost of two calls and that they are the same, in non isolation since the first would be warm the storage it would cost less

DhairyaSethi (approved):

PR #755, opened 2026-07-07
feat: reserve config change helpers

AlbertoCentonze:
Adds reusable Aave V3 reserve config change validation helpers to `ProtocolV3TestBase`.

Tests can now declare expected config-engine changes via override hooks for:

- new standard/custom listings
- collateral parameter updates
- supply and borrow cap updates
- borrow config updates
- freeze state updates

The helper executes the payload, derives the expected post-payload reserve configs from the pre-payload snapshot, validates expected listings/updates were applied, and ensures no undeclared reserve config changes occurred.

Also adds unit coverage for valid expected changes, missing declared changes, unexpected config mutations, and unexpected new listings.

## Checklist

- [ ] Added a [changeset](https://github.com/changesets/changesets) if this modifies a published package (`pnpm changeset`)

PR #761, opened 2026-07-08
feat: reserve config change helpers

AlbertoCentonze:
Adds reusable Aave V3 reserve config change validation helpers to `ProtocolV3TestBase`.

Tests can now declare expected config-engine changes via override hooks for:

- new standard/custom listings
- collateral parameter updates
- supply and borrow cap updates
- borrow config updates
- freeze state updates

The helper executes the payload, derives the expected post-payload reserve configs from the pre-payload snapshot, validates expected listings/updates were applied, and ensures no undeclared reserve config changes occurred.

Also adds unit coverage for valid expected changes, missing declared changes, unexpected config mutations, and unexpected new listings.

Supersedes #755 so CI runs from an upstream branch with access to repository rpc urls.

---
(sorry this scope creeped a bit)

Also annotates `GovernanceV3Test.test_helpers` with `forge-config: default.isolate = true` to fix a `PAYLOAD_GAS_CHECK_REQUIRES_ISOLATION` failure.

#743 added the payload gas-limit check in `CommonTestBase.sol`, which only measures correctly under isolation, and annotated every `defaultTest` caller except `test_helpers`. It stayed green because CI ran on a nightly Foundry with isolate-by-default (foundry#15226), which satisfied the check globally regardless of the annotation. Switching CI to the aave-dao workflows (e3ccbc0), whose Foundry does not isolate by default, is what surfaced the gap.


src/ProtocolV3TestBase.sol
```diff
@@ -271,6 +270,265 @@ contract ProtocolV3TestBase is RawProtocolV3TestBase, SeatbeltUtils, CommonTestB
...
+  }
+
+  function _expectedCustomListings()
+    internal
+    pure
+    virtual
+    returns (IAaveV3ConfigEngine.Listing[] memory listings, uint256[] memory decimals)
+  {
+    listings = new IAaveV3ConfigEngine.Listing[](0);
+    decimals = new uint256[](0);
```
Kogaroshi: those 2 could return a single list with a struct of 2 params, to avoid length mismatch issues

tests/ProtocolV3TestBase.t.sol
```diff
@@ -324,6 +326,282 @@ contract ProtocolV3TestPlausibilityEMode is ProtocolV3TestBase {
   }
 }
 
+contract ProtocolV3TestBaseReserveConfigChangesTest is ProtocolV3TestBase {
```
Kogaroshi, line 329: can we do some fuzz tests with sanitized inputs ?

Kogaroshi: to have more than 1 item array there

src/ProtocolV3TestBase.sol
AlbertoCentonze: I didn't want to create extra structs, length consistency is loosely guaranteed in codegen so I preferred to reduce boilerplate code. Happy to go with the list of struct approach if you think it's more consistent with the current testing approach.

Kogaroshi: Current usage should prevent length mismatch, but future changes might not take it into account correctly

Kogaroshi (approved):

PR #762, opened 2026-07-08
rft: make v4 e2e tests chain agnostic

Kogaroshi:
Make `ProtocolV4TestBase.sol` chain agnostic, and introduce virtual hooks to use aave-address-book getter data for each chain.
A TestBase contract can be found for each chain in `src/v4-protocol-test/` with all the overrides necessary. This designed is favored to keep the tests, input params & imports simpler for Proposal tests in aave-proposals-v3


reports/V3StorageValidation_fail_before.json
DhairyaSethi, line 1: why diff here?

Kogaroshi: getting it diffed every time I do changes in this repo, forgot to revert it back, will do now

DhairyaSethi (approved):

PR #776, opened 2026-08-03
fix: add back removed helpers from proposals repo

Kogaroshi:
Add back some helpers for V4 payload and tests lost during the removal of helpers in aave-proposals-v3


tests/v4-config-engine/V4RoleWiring.t.sol
```diff
@@ -0,0 +1,143 @@
...
+  address internal authority = makeAddr('ACCESS_MANAGER');
+  address internal hub = makeAddr('HUB');
+  address internal spoke = makeAddr('SPOKE');
+
+  function test_hubWiring_rolesAndSelectors() public view {
+    IAaveV4ConfigEngine.TargetFunctionRoleUpdate[] memory updates = V4RoleWiring.hubWiring(
+      authority,
+      hub
+    );
+    assertEq(updates.length, 3, 'updates length');
```
avniculae: These asserts can't catch the failure mode the library exists to prevent: both sides come from `Roles` (`_assertUpdate(..., Roles.HUB_CONFIGURATOR_ROLE, Roles.getHubConfiguratorRoleSelectors())`) and the counts are hardcoded 3 and 2. If aave-v4 adds a fourth hub role to `setupHubAllRoles`, `hubWiring` keeps returning three and everything here stays green — which is how the spoke 302 gap and then the hub 102/103 gap happened in the first place.

Would an equivalence test against the deployment procedure be worth adding? `vm.recordLogs`, run `AaveV4HubRolesProcedure.setupHubAllRoles(am, hub)` against a real AccessManager, then assert `hubWiring(am, hub)` produces exactly the same set of (target, selector, roleId) triples. Both procedures are internal libraries reachable through the existing `aave-v4/` remap, so it should be short, and it would fail loudly on an upstream addition instead of silently under-wiring.

src/ProtocolV4TestBase.sol
```diff
@@ -202,6 +202,29 @@ abstract contract ProtocolV4TestBase is
     }
   }
 
+  /// @notice Assert a payload's declared role wiring landed on its AccessManager. Call after execution.
+  /// @param item One entry of the payload's `accessManagerTargetFunctionRoleUpdates()`.
+  /// @param referenceTarget Already-wired contract asserted to carry the same role per selector,
+  ///        or `address(0)` to skip that check.
+  function _assertRolesWired(
```
avniculae, line 211: This is the one added symbol with no test, and it's the helper payload tests will actually depend on for their role assertions. A positive path over a mock AccessManager plus the `referenceTarget == address(0)` skip would cover most of it.

src/dependencies/v4/Helpers.sol
```diff
@@ -61,6 +66,28 @@ abstract contract Helpers is Actions {
...
+    uint256 assetId = hub.getAssetId(underlying);
+    for (uint256 i = hub.getSpokeCount(assetId); i > 0; --i) {
+      address spoke = hub.getSpokeAddress(assetId, i - 1);
+      try ITokenizationSpoke(spoke).asset() returns (address tokenized) {
+        if (tokenized == underlying) {
+          return spoke;
+        }
+      } catch {}
+    }
+    return address(0);
```
avniculae, line 82: Returning `address(0)` here differs from the copy the generator emits in aave-proposals-v3#1177, which reverts with 'tokenization spoke not found'. Two implementations with different not-found semantics, and a caller that skips the zero check gets an obscure revert out of `Ownable(address(0)).owner()` in `_proxyAdminOwner` rather than a clear message.

Worth settling on one semantic and having the generator call this helper instead of emitting its own `_findTokenizationSpoke_<hub>_<asset>`? Otherwise the duplication outlives this PR.

src/v4-config-engine/V4EngineDefaults.sol
```diff
@@ -0,0 +1,49 @@
...
+  {
+    return
+      IAaveV4ConfigEngine.TokenizationSpokeConfig({
+        addCap: 0,
+        proxyAdminOwner: address(0),
+        name: '',
+        symbol: ''
+      });
+  }
+}
```
avniculae, line 49: Two things from the removed set didn't come back, both cheap:

The tokenization name/symbol convention (`'Wrapped Aave ' + hubName + ' ' + token` / `'wa' + hubName + token`). Not about the Paxos → Global Dollar rename, which is a deliberate change — just that the string format is currently retyped by hand in each payload, and a builder here would keep the shape consistent.

The spoke-deployment sanity checks the old fork base had: `authority() == AccessManager`, `ORACLE()` wired with `oracle.spoke() == spoke` and 8 decimals, and code exists at the spoke. #1175 shipped without them and those ended up being checked by hand on-chain instead.

Kogaroshi: > The tokenization name/symbol convention ('Wrapped Aave ' + hubName + ' ' + token / 'wa' + hubName + token). Not about the Paxos → Global Dollar rename, which is a deliberate change — just that the string format is currently retyped by hand in each payload, and a builder here would keep the shape consistent.

for this point, I plan to add it in the generator changes coming up right after this PR.

Kogaroshi: for the 2nd one, I can definitely see to add it back yes.

Kogaroshi: 6d633f26813eeb592558288d2c71fc5eccbf74ce

src/ProtocolV4TestBase.sol
Kogaroshi: 6d633f26813eeb592558288d2c71fc5eccbf74ce

tests/v4-config-engine/V4RoleWiring.t.sol
Kogaroshi: 6d633f26813eeb592558288d2c71fc5eccbf74ce

src/dependencies/v4/Helpers.sol
Kogaroshi: The goal here is to keep the current `return address(0)` behavior here, so usage of that helper is not always reverting if not found. And in aave-proposals-v3, this will be extend into a revert cases in the template tests for the Payload Generator.

reports/V3StorageValidation_fail_before.json
```diff
@@ -1,2965 +1,3 @@
...
-    }
-  },
-  "poolConfig": {
-    "oracle": "0x54586bE62E3c3580375aE3723C145253060Ca0C2",
-    "pool": "0x87870Bca3F3fD6335C3F4ce8392D69350B4fA4E2",
-    "poolAddressesProvider": "0x2f39d218133AFaB8F2B819B1066c7E434Ad94E9e",
-    "poolConfigurator": "0x64b761D848206f447Fe2dd461b0c635Ec39EbB27",
-    "protocolDataProvider": "0x0a16f2FCC0D44FaE41cc54e079281D84A363bECD"
-  }
+  "chainId": 1
```
avniculae: Is this emptying intentional? Every reserve is gone here (2964 lines -> `{"chainId": 1}`), while the siblings are untouched — `V3StorageValidation_fail_after.json` is still 2964 and `V3StorageValidation_pass_before.json` still 2964.

The test writing it (`tests/ProtocolV3TestBase.t.sol:749`, against `AaveV3Ethereum.POOL`) isn't modified by this PR, and an empty before against a full after would make that diff read as if every Ethereum reserve were newly added, which isn't what the fail case is meant to exercise. Looks more like a stray local run got committed — and reports are outputs rather than assertions, so CI wouldn't flag it.

The `diffs/mantle_*` and `diffs/megaeth_*` churn in the same commit does look expected, just label re-resolution from the address-book bump picking up MegaEth's new USDe/stcUSD entries.

Kogaroshi: No, this is a leftover from me running again the full test suite, didn't mean to touch V3 side, I'll revert those

Kogaroshi: d7de2bc593ff2c517cc00d2b6a86faf37c4bb008

avniculae (approved):

PR #783, opened 2026-08-17
feat: add GSM events to eventDb

efecarranza:
Adds 12 missing events from the GSM implementation on Arbitrum (0x53E0cE250d06043414070100458546AaF4e284eD, impl 0x66c6ba44...fd855) so GSM transaction logs decode.

### DIFF
--- a/packages/aave-helpers-js/utils/eventDb.ts
+++ b/packages/aave-helpers-js/utils/eventDb.ts
@@ -3804,4 +3804,64 @@ export const eventDb: AbiEvent[] = [
     name: 'TokensRescued',
     type: 'event',
   },
+  {
+    anonymous: false,
+    inputs: [{ indexed: true, internalType: 'address', name: 'entity', type: 'address' }],
+    name: 'EntityAdded',
+    type: 'event',
+  },
+  {
+    anonymous: false,
+    inputs: [{ indexed: true, internalType: 'address', name: 'entity', type: 'address' }],
+    name: 'EntityRemoved',
+    type: 'event',
+  },
+  {
+    anonymous: false,
+    inputs: [
+      { indexed: true, internalType: 'address', name: 'entity', type: 'address' },
+      { indexed: false, internalType: 'uint256', name: 'limit', type: 'uint256' },
+    ],
+    name: 'GhoLimitUpdated',
+    type: 'event',
+  },
+  {
+    anonymous: false,
+    inputs: [
+      { indexed: true, internalType: 'address', name: 'entity', type: 'address' },
+      { indexed: false, internalType: 'uint256', name: 'amount', type: 'uint256' },
+    ],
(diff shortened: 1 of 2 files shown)


DhairyaSethi (dismissed):
lgtm pls resolve conflicts

packages/aave-helpers-js/utils/eventDb.ts
```diff
@@ -3682,4 +3682,126 @@ export const eventDb: AbiEvent[] = [
...
+  {
+    anonymous: false,
+    inputs: [
+      { indexed: true, internalType: 'address', name: 'tokenRescued', type: 'address' },
+      { indexed: true, internalType: 'address', name: 'recipient', type: 'address' },
+      { indexed: false, internalType: 'uint256', name: 'amountRescued', type: 'uint256' },
+    ],
+    name: 'TokensRescued',
+    type: 'event',
+  },
```
avniculae, line 3944: Checked all 12 against gho-origin `main` (names, param types, order, indexed flags) and they match, including the ones that are easy to get wrong: `GhoReserveUpdated` with both params non-indexed, `ExposureCapUpdated` with none indexed. topic0 of `ExposureCapUpdated` / `Seized` / `FeeStrategyUpdated` also matches the raw topics in the Arbitrum GSM replacement diff report, so those rows will decode.

The report has five more undecoded rows that this PR doesn't cover though, since they're emitted by `GhoReserve` and `GsmRegistry` rather than the GSM itself:

| topic0 | event | contract |
| --- | --- | --- |
| `0xb22e19f0...` | `EntityAdded(address indexed entity)` | GhoReserve |
| `0x66baa859...` | `EntityRemoved(address indexed entity)` | GhoReserve |
| `0x29d76d87...` | `GhoLimitUpdated(address indexed entity, uint256 limit)` | GhoReserve |
| `0x3e31b6b8...` | `GsmAdded(address indexed gsmAddress)` | GsmRegistry |
| `0x207b84f0...` | `GsmRemoved(address indexed gsmAddress)` | GsmRegistry |

Could you add those here as well? Otherwise the regenerated report still prints raw topics for half of its rows. While you're at it, `GhoUsed(address indexed entity, uint256 amount)` and `GhoRestored(address indexed entity, uint256 amount)` are worth including, since any report covering a GSM swap will emit them.

avniculae:
Unrelated to this PR, but while checking for collisions I noticed eventDb on main already has two exact duplicates: `ERC20Rescued` (lines 1955 and 2213) and `NativeTokensRescued` (lines 1965 and 2252), same signature and indexed flags both times. Harmless for decoding since matching is by topic0. Tracked in #798 for a small cleanup PR.

avniculae:
Superseded by #802: same commits re-raised from a repo branch so we can look at the CI failure separately. Closing this one.

Issue #813, opened 2026-09-19
Gas-cap the symbol() probe in Helpers._safeSymbol


DhairyaSethi:
`Helpers._safeSymbol` forwards all remaining gas to `token.staticcall(symbol())`. When the token's code cannot be executed by the EVM the call burns everything it was given, and since each probe gets 63/64 of the caller's gas, a snapshot over several such tokens starves the test of gas and it dies with a bare `EvmError: Revert` instead of rendering `<unknown>`.

Concrete case: the Base V4 equities market lists seven B20 tokens whose account code is the single byte `0xef`. Under stock forge every call to them reverts this way, and `test_defaultProposalExecution` on the Base activation proposal fails before reaching the diff.

Suggested fix, applied on branch `feat/base-v4`:

```solidity
(bool ok, bytes memory data) = token.staticcall{gas: 100_000}(
  abi.encodeCall(IERC20Metadata.symbol, ())
);
```

`symbol()` on a normal token costs a few thousand gas, so the cap only changes behaviour for calls that were going to fail. A token whose `symbol()` needs more than the cap renders as `<unknown>`, which is the existing fallback. The same cap is worth applying to any other metadata probe in the snapshot path.

PR #814, opened 2026-09-19
feat: Base V4 bindings and B20 equity deal path

DhairyaSethi:
Adds the test base and payload base for the Aave V4 market on Base, pins aave-dao/aave-address-book at the 4.69.0 release commit 8d51d3ed, and funds the seven B20 equity tokens in `_patchedDeal` by minting from the `MINT_ROLE` holder, since their balances live outside EVM storage and `deal` cannot find a slot. `Helpers._safeSymbol` now caps the gas it forwards so that a token the EVM cannot execute renders as `<unknown>` instead of starving the snapshot (#813).

### Notes

Pins aave-dao/aave-address-book at the 4.69.0 release commit 8d51d3ed; the npm publish of 4.69.0 was still pending at the time of writing. The B20 mint path only executes under base-anvil's forge (`FOUNDRY_BASE=true`); under stock forge every call to those tokens reverts, which is what the gas cap absorbs.


src/CommonTestBase.sol
```diff
@@ -186,6 +188,23 @@ contract CommonTestBase is Test {
...
+      // find a slot. Mint from the MINT_ROLE holder instead; only executable under base-anvil's forge.
+      if (
+        asset == AaveV4BaseAssets.AAPLc_UNDERLYING ||
+        asset == AaveV4BaseAssets.AMZNc_UNDERLYING ||
+        asset == AaveV4BaseAssets.GOOGLc_UNDERLYING ||
+        asset == AaveV4BaseAssets.METAc_UNDERLYING ||
+        asset == AaveV4BaseAssets.MSFTc_UNDERLYING ||
+        asset == AaveV4BaseAssets.NVDAc_UNDERLYING ||
+        asset == AaveV4BaseAssets.TSLAc_UNDERLYING
+      ) {
```
Kogaroshi: let's maybe add a library file with those addresses and a getter to check if the asset is a B20 ?
or the other solution might be to recognize any address with its code being `0xef` as a B20, even if listed from our address-book or not ? 
Just thinking long term, how we'd need to update that list for each new B20 listing.

DhairyaSethi: thinking we'll check with factory bc `0xef` can be mis understood for another pre compile

Kogaroshi (dismissed):

src/CommonTestBase.sol
DhairyaSethi: Done in 2d55b1f: `_patchedDeal` now asks the factory precompile `isB20`/`isB20Initialized` and drops the address list; under stock forge the factory account is empty so the call returns no data and normal `deal` proceeds. The mint path asserts the supply manager holds `MINT_ROLE` and reverts naming the token otherwise; fork tests cover AAPLc (base-forge) and USDC (both toolchains).

DhairyaSethi: Reverted from this PR in f1d801f so #814 merges with the hardcoded list; the factory-based detection moved to https://github.com/aave-dao/aave-helpers/pull/817.

Kogaroshi (approved):

PR #817, opened 2026-09-21
feat: detect Base B20 tokens through the factory

### LINKED ISSUE #816: Detect Base B20 tokens through the factory instead of a hardcoded address list
`CommonTestBase._patchedDeal` (branch `feat/base-v4`, #814) recognises the seven Coinbase equity tokens on Base by comparing against the `AaveV4BaseAssets.*_UNDERLYING` constants before minting from the `MINT_ROLE` holder. Any B20 token listed later on Base falls through to `deal`, which cannot find a storage slot for a precompile-backed balance and fails. Checking for account code `0xef` is not a substitute: other Base precompiles carry the same byte.

The B20 factory precompile at `0xB20f000000000000000000000000000000000000` exposes the right predicate (see `IB20Factory` in base/base-std):

```solidity
/// @notice Returns whether `token` was created by this factory, recovered from the address prefix. Never reverts.
function isB20(address token) external view returns (bool);
/// @notice Returns whether `createB20` has run to completion at `token`. Never reverts.
function isB20Initialized(address token) external view returns (bool);
```

Suggested change: on `ChainIds.BASE`, call `IB20Factory(0xB20f…0000).isB20(asset)` (and `isB20Initialized`) and, when true, mint from the token's `MINT_ROLE` holder; drop the address list. The mint-role holder still needs a source. For the Coinbase tokens it is the supply manager `0xD1Ca4dAcdf3231011D175351f1f02D15C7c5664C`; a generic path could read `RoleGranted(MINT_ROLE, …)` logs or accept a per-token override, otherwise fall back to the current constant with a clear revert message when the token is a B20 without a known minter.

Raised from the review comment https://github.com/aave-dao/aave-helpers/pull/814#discussion_r4062855307.

DhairyaSethi:
`CommonTestBase._patchedDeal` now asks the B20 factory precompile (`0xB20f000000000000000000000000000000000000`) `isB20` and `isB20Initialized` instead of comparing against the seven hardcoded `AaveV4BaseAssets.*_UNDERLYING` constants, so any B20 listed later on Base takes the mint path. Before minting it asserts that the Coinbase supply manager holds `MINT_ROLE` on the token and reverts naming the token otherwise. Base fork tests (block 51605828) cover detection and mint of AAPLc, the mint-role revert, USDC taking the normal `deal` path, and all seven known equities being initialised B20s mintable by the supply manager.

### Notes

- Stacks on #814; GitHub retargets to `main` once that merges.
- Closes #816.
- The B20 tests skip with a stated reason under stock forge because neither the factory precompile nor the 0xef tokens execute there; they run under base-anvil's forge. The USDC test runs on both toolchains.


tests/CommonTestBase.t.sol
```diff
@@ -33,3 +36,71 @@ contract CommonTestBaseGnosisTest is CommonTestBase {
     assertEq(IERC20(AaveV3GnosisAssets.EURe_UNDERLYING).balanceOf(address(this)), 100e18);
   }
 }
+
+contract CommonTestBaseBaseTest is CommonTestBase {
```
Kogaroshi, line 40: Idk about this name, but can't find a batter one ig

Kogaroshi (dismissed):

Kogaroshi (dismissed):

Kogaroshi (approved):

